PR #21, opened 2023-11-09
Ignore test code in coverage reports, integrate with codecov

tomleavy:
Generate a clean coverage report by eliminating code that is only compiled under the test cfg. Add support for codecov under github actions

### DIFF
--- a/mls-rs/src/group/ciphertext_processor/reuse_guard.rs
+++ b/mls-rs/src/group/ciphertext_processor/reuse_guard.rs
@@ -92,23 +92,27 @@ mod tests {
         result: Vec<u8>,
     }
 
+    #[cfg_attr(coverage_nightly, coverage(off))]
     fn generate_reuse_guard_test_cases() -> Vec<TestCase> {
         let provider = test_cipher_suite_provider(TEST_CIPHER_SUITE);
 
         [16, 32]
             .into_iter()
-            .map(|len| {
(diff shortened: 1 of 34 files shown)


mulmarta (approved):

mls-rs/src/group/ciphertext_processor/reuse_guard.rs
```diff
@@ -92,23 +92,27 @@ mod tests {
...
-                let result = guard.apply(&nonce);
-
-                TestCase {
-                    nonce,
-                    guard: guard.into(),
-                    result,
-                }
-            })
+            .map(
+                #[cfg_attr(coverage_nightly, coverage(off))]
```
stefunctional, line 102: Why is it necessary to place that on the lambda definition when the containing function already has `coverage(off)`?

stefunctional (approved):

mls-rs/src/group/ciphertext_processor/reuse_guard.rs
tomleavy: It won't ignore it otherwise. Pretty sure this is a nuance of LLVM where those lambdas are internally converted into anonymous functions

stefunctional: Thank you for explaining.

PR #22, opened 2023-11-09
Fix example

mulmarta:
The example needs update paths for add-only commits, in order to create full tree.

### DIFF
--- a/mls-rs/Cargo.toml
+++ b/mls-rs/Cargo.toml
@@ -111,7 +111,7 @@ required-features = ["x509"]
 
 [[example]]
 name = "large_group"
-required-features = []
+required-features = ["test_util"]
 
 [[example]]
 name = "basic_server_usage"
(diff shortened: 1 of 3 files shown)


tomleavy (dismissed):

mls-rs/Cargo.toml
```diff
@@ -111,7 +111,7 @@ required-features = ["x509"]
 
 [[example]]
 name = "large_group"
-required-features = []
+required-features = ["test_util"]
```
stefunctional, line 114: That's a little odd to have to enable a test-related feature for an example. I don't remember why `DefaultMlsRules` is defined differently in test vs no-test.

tomleavy: Agreed we shouldn't require test_util for that

mulmarta: I made the customizable ones default, since it's handy and doesn't hurt.

stefunctional (approved):

tomleavy (approved):

PR #23, opened 2023-11-09
Restore HPKE tests

### LINKED ISSUE #19: Adjust HPKE tests in mls-rs-core
We had a set of duplicate tests, one in HPKE and another in core which is used to determine if a crypto provider is meeting the standard properly. We should replace the test vector data in the currently active test within mls-rs-core with the one that was previously used in mls-rs-hpke

mulmarta:
### Issues:

Resolves #19 

### Description of changes:

Restore the HPKE tests from RFC 9180 (JSON without change). The tests are more low-level than the CryptoProvider API (fixed ephemeral key etc), so they're exposed as additional functions in `core::test_suite`.

* To run the test, one has to implement the `TestHpke` trait for their HPKE.
* `mls-rs-core-hpke` has a "test_utils" feature which implements the trait.

### DIFF
--- a/mls-rs-core/src/crypto/cipher_suite.rs
+++ b/mls-rs-core/src/crypto/cipher_suite.rs
@@ -81,3 +81,17 @@ impl CipherSuite {
         (1..=7).map(CipherSuite)
     }
 }
+
+/// Modes of HPKE operation.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
+pub enum HpkeModeId {
+    /// Base mode of HPKE for key exchange and AEAD cipher
+    Base = 0x00,
+    /// Base mode with a user provided PSK
+    Psk = 0x01,
+    /// Authenticated variant that authenticates possession of a KEM private key.
+    Auth = 0x02,
+    /// Authenticated variant that authenticates possession of a PSK as well as a KEM private key.
+    AuthPsk = 0x03,
+}
--- a/mls-rs-crypto-hpke/Cargo.toml
+++ b/mls-rs-crypto-hpke/Cargo.toml
@@ -10,7 +10,7 @@ categories = ["no-std", "cryptography"]
 license = "Apache-2.0 OR MIT"
 
 [features]
-default = ["std", "test_utils"]
+default = ["std"]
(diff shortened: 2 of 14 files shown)


mls-rs-crypto-hpke/Cargo.toml
```diff
@@ -10,8 +10,9 @@ categories = ["no-std", "cryptography"]
 license = "Apache-2.0 OR MIT"
 
 [features]
-default = ["std"]
```
tomleavy, line 13: I don't think `test_utils` should be part of the default here?

mls-rs-crypto-hpke/Cargo.toml
```diff
@@ -10,8 +10,9 @@ categories = ["no-std", "cryptography"]
...
+default = ["std", "test_utils"]
```
tomleavy, line 13: I don't think `test_utils` should be part of the default here?

mulmarta, line 13:
```suggestion
default = ["std"]
```

Leftover debug...

tomleavy (approved):

mls-rs-core/src/crypto/cipher_suite.rs
```diff
@@ -81,3 +81,17 @@ impl CipherSuite {
         (1..=7).map(CipherSuite)
     }
 }
+
+/// Modes of HPKE operation.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
+pub enum HpkeModeId {
```
stefunctional, line 88: Should it be `#[non_exhaustive]` or are we sure new variants will never be added?

stefunctional (approved):

mls-rs-core/src/crypto/cipher_suite.rs
mulmarta: They are defined in the [RFC table](https://www.rfc-editor.org/rfc/rfc9180.html#table-1) and they don't seem to be extensible but I don't mind adding the flag. There may be more such places, I can double check in a follow up PR?

tomleavy: Lets do a follow up

PR #34, opened 2023-11-13
Use `ProposalInfo` instead of `Proposal` for unused proposals

### LINKED ISSUE #27: Unused proposals in StateUpdate should have more information
### Description of feature:

We should return the ProposalRef as well so that you can easily look them up

### Use case:

You might have proposals stored somewhere else with the ProposalRef as a unique lookup key

tomleavy:
### Issues:

Resolves #27 

### Description of changes:

Replace `Proposal` with `ProposalInfo<Proposal>` to offer a better description of the proposal that was not used.

### Call-outs:

Chose this over `CachedProposal` since this has the same information and is used in the public API already.

### Testing:

Updated a lot of tests to verify. 

### DIFF
--- a/mls-rs/src/group/proposal_cache.rs
+++ b/mls-rs/src/group/proposal_cache.rs
@@ -176,7 +176,6 @@ impl GroupState {
     pub(crate) async fn apply_resolved<C, F, P, CSP>(
         &self,
         sender: Sender,
-        #[cfg(all(feature = "by_ref_proposal", feature = "state_update"))] receiver: Option<Sender>,
         mut proposals: ProposalBundle,
         external_leaf: Option<&LeafNode>,
         identity_provider: &C,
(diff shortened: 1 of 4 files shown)


mulmarta (approved):

mls-rs/src/group/proposal_cache.rs
```diff
@@ -299,17 +295,12 @@ fn has_ref(proposals: &ProposalBundle, reference: &ProposalRef) -> bool {
...
         .into_proposals()
-        .filter_map(|p| match p.source {
-            ProposalSource::ByReference(r)
-                if !has_ref(accepted_proposals, &r) && &p.sender == sender =>
-            {
-                Some((r, p.proposal))
-            }
-            _ => None,
+        .filter(|p| {
+            matches!(p.source, ProposalSource::ByReference(ref r) if !has_ref(accepted_proposals, r)
```
stefunctional, line 302: It seems that rejected proposals will now contain proposals from all members instead of only the one sending or receiving the commit? Is that an intentional change of behavior?

stefunctional (approved):

mls-rs/src/group/proposal_cache.rs
mulmarta: Yes, and that's what I like most about this PR ;-)

tomleavy: Yeah, that is required in order to use this within the `ExternalClient` and it makes sense because you can always filter it post receiving this information.

PR #41, opened 2023-11-16
Improve interop tests CI

mulmarta:
The tests are run sequentially so that 
1. we see how long each test takes
2. there are at most 2 processes ever

### DIFF
--- a/.github/workflows/interop_tests.yml
+++ b/.github/workflows/interop_tests.yml
@@ -14,32 +14,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           path: mls
-      - name: Start test MLS client
-        working-directory: mls-rs/test_harness_integration
-        run: |
-          cargo run -- --port 50001 &
-          cargo run --no-default-features -- --port 50002 &
-
-          # no tree_index
-          cargo run --no-default-features --features private_message,prior_epoch,out_of_order,psk,custom_proposal,by_ref_proposal -- --port 50003 &
-
-          # no private_message
-          cargo run --no-default-features --features tree_index,prior_epoch,out_of_order,psk,custom_proposal,by_ref_proposal -- --port 50004 &
-
-          # no prior_epoch
-          cargo run --no-default-features --features tree_index,private_message,out_of_order,psk,custom_proposal,by_ref_proposal -- --port 50005 &
-
-          # no out_of_order
-          cargo run --no-default-features --features tree_index,private_message,prior_epoch,psk,custom_proposal,by_ref_proposal -- --port 50006 &
-
-          # no psk
-          cargo run --no-default-features --features tree_index,private_message,prior_epoch,out_of_order,custom_proposal,by_ref_proposal -- --port 50007 &
-
-          # no custom_proposal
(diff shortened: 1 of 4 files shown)


.github/workflows/interop_tests.yml
```diff
@@ -62,30 +36,92 @@ jobs:
...
-
-           # all features with "no by_ref_proposal"
-          for config in `ls mls-rs-configs | grep -E "(application)|(commit_by_value)|(branch)|(welcome_join)" | sed -e "s/mls-rs-configs\///"`; do >&2 echo $config && test-runner/test-runner --client localhost:50001 --client localhost:50009 --suite 1 --config mls-rs-configs/$config ; done > /dev/null
+          kill %1
+          kill %2
+      - name: Test interop; full feature with "no tree_index"
+        run: |
+          cd mls-rs/test_harness_integration
+          cargo run -- --port 50001 &
+          cargo build --no-default-features --features private_message,prior_epoch,out_of_order,psk,custom_proposal,by_ref_proposal
```
stefunctional, line 61: Building before running should not be necessary, as `cargo run` triggers a build if the artifacts are out of date.

mls-rs/src/group/message_processor.rs
```diff
@@ -75,7 +75,7 @@ pub(crate) struct ProvisionalState {
     pub(crate) external_init_index: Option<LeafIndex>,
     pub(crate) indexes_of_added_kpkgs: Vec<LeafIndex>,
     #[cfg(all(feature = "state_update", feature = "by_ref_proposal"))]
-    pub(crate) rejected_proposals: Vec<ProposalInfo<Proposal>>,
+    pub(crate) rejected_proposals: Vec<crate::mls_rules::ProposalInfo<Proposal>>,
```
stefunctional, line 78: Out of curiosity, why is the path fully qualified now?

stefunctional (approved):

.github/workflows/interop_tests.yml
mulmarta: It sure does, but `cargo run &` in the background doesn't return the exit code until you `wait` for the process. So, it does `cargo build` to get the exit code and kill the action on fail. Does this make sense?

mls-rs/src/group/message_processor.rs
mulmarta: Because in a weird feature combination (the interop runner uses different ones than `cargo hack`) the import was not compiled. It was too difficult to figure out when to import :/

stefunctional: Thank you for explaining.

.github/workflows/interop_tests.yml
stefunctional: So is the build done first to ensure the CI job fails immediately if a build error occurs?

mulmarta: Yes, exactly

stefunctional: Makes sense, thank you. And given that it seems to start servers and clients, I guess you may run into connection issues if the server were to take too long to start because it needs to be built first.

tomleavy (approved):

PR #54, opened 2023-12-07
Expose tree hash

mulmarta:


tomleavy (approved):

mls-rs/src/external_client/group.rs
```diff
@@ -518,6 +518,14 @@ impl<C: ExternalClientConfig + Clone> ExternalGroup<C> {
         &self.group_state().context.confirmed_transcript_hash
     }
 
+    /// Get the
+    /// [tree hash](https://www.rfc-editor.org/rfc/rfc9420.html#name-tree-hashes)
+    /// for the current epoch that the group is in.
+    #[inline(always)]
+    pub fn tree_hash(&self) -> &Vec<u8> {
```
stefunctional, line 525: [Nit] `&[u8]` is almost always returned instead of `&Vec<u8>`, because the latter does not provide any additional useful functionality, but it has stricter requirements (the backend storage must be a full `Vec`, so it can't be `Box<u8>`, part of a `Vec` or array, or some other structure). It is unlikely to ever change from a `Vec` here, but I thought I'd mention it anyway.

```suggestion
    pub fn tree_hash(&self) -> &[u8] {
```

stefunctional (dismissed):

PR #58, opened 2023-12-20
Introduce `ExportedTree` to represent ratchet tree contents

tomleavy:
### Description of changes:

Currently the ratchet tree is exported as a `Vec<u8>` and imported as `Option<&[u8]>`. It is more flexible to make this a wrapper around `NodeVec` so that serialization operations can happen separately.

### DIFF
--- a/mls-rs-codec/src/cow.rs
+++ b/mls-rs-codec/src/cow.rs
@@ -1,4 +1,7 @@
-use alloc::borrow::Cow;
+use alloc::{
+    borrow::{Cow, ToOwned},
+    vec::Vec,
+};
 
 use crate::{Error, MlsDecode, MlsEncode, MlsSize};
 
--- a/mls-rs-codec/src/cow.rs
+++ b/mls-rs-codec/src/cow.rs
@@ -1,0 +1,31 @@
+use alloc::borrow::Cow;
+
+use crate::{Error, MlsDecode, MlsEncode, MlsSize};
+
+impl<'a, T> MlsSize for Cow<'a, T>
+where
+    T: MlsSize + ToOwned,
+{
+    fn mls_encoded_len(&self) -> usize {
+        self.as_ref().mls_encoded_len()
+    }
+}
+
+impl<'a, T> MlsEncode for Cow<'a, T>
+where
+    T: MlsEncode + ToOwned,
+{
+    #[inline]
+    fn mls_encode(&self, writer: &mut Vec<u8>) -> Result<(), Error> {
+        self.as_ref().mls_encode(writer)
+    }
+}
+
+impl<'a, T> MlsDecode for Cow<'a, T>
+where
+    T: MlsDecode + ToOwned<Owned = T>,
+{
+    fn mls_decode(reader: &mut &[u8]) -> Result<Self, Error> {
+        T::mls_decode(reader).map(Cow::Owned)
+    }
+}
--- a/mls-rs-codec/src/cow.rs
+++ b/mls-rs-codec/src/cow.rs
@@ -26,9 +26,10 @@ where
 
 impl<'a, T> MlsDecode for Cow<'a, T>
 where
-    T: MlsDecode + ToOwned<Owned = T>,
+    T: ToOwned,
+    <T as ToOwned>::Owned: MlsDecode,
 {
     fn mls_decode(reader: &mut &[u8]) -> Result<Self, Error> {
-        T::mls_decode(reader).map(Cow::Owned)
+        MlsDecode::mls_decode(reader).map(Cow::Owned)
     }
 }
--- a/mls-rs/examples/basic_server_usage.rs
+++ b/mls-rs/examples/basic_server_usage.rs
@@ -39,10 +39,9 @@ struct BasicServer {
 
 impl BasicServer {
     // Client uploads group data after creating the group
-    fn create_group(group_info: &[u8], tree: &[u8]) -> Result<Self, MlsError> {
+    fn create_group(group_info: &[u8], tree: ExportedTree) -> Result<Self, MlsError> {
         let server = make_server();
         let group_info = MlsMessage::from_bytes(group_info)?;
-        let tree = ExportedTree::from_bytes(tree)?;
 
         let group = server.observe_group(group_info, Some(tree))?;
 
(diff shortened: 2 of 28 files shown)


mls-rs/src/group/mod.rs
```diff
@@ -1434,11 +1438,8 @@ where
...
     /// This function is used to provide the current group tree to new members
     /// when the `ratchet_tree_extension` is not used according to [`MlsRules::commit_options`].
-    pub fn export_tree(&self) -> Result<Vec<u8>, MlsError> {
-        self.current_epoch_tree()
-            .export_node_data()
-            .mls_encode_to_vec()
-            .map_err(Into::into)
+    pub fn export_tree(&self) -> ExportedTree {
+        ExportedTree::new(self.current_epoch_tree().export_node_data())
     }
```
mulmarta, line 1443: This was already there before, but is cloning the tree necessary here? Exported `ExportedTree` could have a borrow of the nodes, while imported `ExportedTree` could have a copy (can be different types).

mls-rs/src/group/exported_tree.rs
```diff
@@ -0,0 +1,32 @@
...
+        Self(node_data)
+    }
+
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, MlsError> {
+        Self::mls_decode(&mut &*bytes).map_err(Into::into)
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, MlsError> {
+        self.mls_encode_to_vec().map_err(Into::into)
+    }
```
mulmarta, line 25: and also `pub fn byte_size(&self) -> usize` for convenience?

mls-rs/src/group/commit.rs
```diff
@@ -674,13 +675,7 @@ where
...
-            .then(|| {
-                provisional_state
-                    .public_tree
-                    .export_node_data()
-                    .mls_encode_to_vec()
-            })
-            .transpose()?;
+            .then(|| ExportedTree::new(provisional_state.public_tree.export_node_data()));
 
         if let Some(signer) = new_signer {
```
mulmarta, line 680: This clones the tree while it could move it out, since `provisional_state` is later deleted.

mls-rs/examples/basic_server_usage.rs
```diff
@@ -155,7 +157,7 @@ fn main() -> Result<(), MlsError> {
 
     // Server starts observing Alice's group
     let group_info = alice_group.group_info_message(true)?.to_bytes()?;
-    let tree = alice_group.export_tree()?;
+    let tree = alice_group.export_tree().to_bytes().unwrap();
```
mulmarta, line 160: [nit] this function generally doesn't panic

mls-rs/examples/basic_server_usage.rs
```diff
@@ -42,6 +42,8 @@ impl BasicServer {
     fn create_group(group_info: &[u8], tree: &[u8]) -> Result<Self, MlsError> {
         let server = make_server();
         let group_info = MlsMessage::from_bytes(group_info)?;
+        let tree = ExportedTree::from_bytes(tree).unwrap();
+
         let group = server.observe_group(group_info, Some(tree))?;
```
mulmarta, line 47: As below, this function doesn't panic

mls-rs/src/group/mod.rs
stefunctional: An alternative to having two types is to define `ExportedTree` as follows:
```rust
struct ExportedTree<'a>(Cow<'a, NodeVec>);
```
The owned case is then `ExportedTree<'static>`.

stefunctional (approved):

tomleavy:
@mulmarta @stefunctional I pushed a change here that makes things based on Cow, but as you can see, FFI is having an issue with that and I'm not sure how to resolve. It also requires that you explicitly call into_owned in a few scenarios. Let me know if you like this (and can fix FFI) or if I should revert and just keep the other comments in the thread

mulmarta:
> @mulmarta @stefunctional I pushed a change here that makes things based on Cow, but as you can see, FFI is having an issue with that and I'm not sure how to resolve. It also requires that you explicitly call into_owned in a few scenarios. Let me know if you like this (and can fix FFI) or if I should revert and just keep the other comments in the thread

@tomleavy I can fix FFI and test harness and an example that also don't build. The thing is
```
impl ExportedTree<'static> {
    pub fn from_bytes(bytes: &[u8]) -> Result<Self, MlsError> {
        Self::mls_decode(&mut &*bytes).map_err(Into::into)
    }
}
```
With `'static` instead of `'a`. I can push to the PR is you wish, or I attached the diff 
[0001-Fix-build.patch](https://github.com/awslabs/mls-rs/files/13777005/0001-Fix-build.patch)


To me these `into_owned` instances make sense since we want to clone.

What do you think @stefunctional ?

tomleavy:
Ah that makes sense, thanks @mulmarta

mls-rs-codec/src/cow.rs
```diff
@@ -0,0 +1,31 @@
...
+{
+    #[inline]
+    fn mls_encode(&self, writer: &mut Vec<u8>) -> Result<(), Error> {
+        self.as_ref().mls_encode(writer)
+    }
+}
+
+impl<'a, T> MlsDecode for Cow<'a, T>
+where
+    T: MlsDecode + ToOwned<Owned = T>,
```
stefunctional, line 26: I think the proper bound should be:
```suggestion
    T: ToOwned,
    <T as ToOwned>::Owned: MlsDecode,
```

This way this implementation also applies to types like `Cow<'_, str>` where the owned and borrowed types are different. We only care about the owned type being decodable here.

mls-rs-codec/src/cow.rs
```diff
@@ -0,0 +1,31 @@
...
+{
+    fn mls_decode(reader: &mut &[u8]) -> Result<Self, Error> {
+        T::mls_decode(reader).map(Cow::Owned)
```
stefunctional, line 29: If applying the above suggestion, this needs to change as follows:
```suggestion
        MlsDecode::mls_decode(reader).map(Cow::Owned)
```

stefunctional (approved):

mulmarta (approved):

stefunctional (dismissed):

stefunctional (approved):

PR #60, opened 2024-01-05
Expose a function to compute `ProposalRef` from a proposal `MlsMessage`

mulmarta:

### DIFF
--- a/mls-rs/src/group/framing.rs
+++ b/mls-rs/src/group/framing.rs
@@ -622,7 +622,7 @@ mod tests {
             }),
         };
 
-        let computed_ref = test_message.proposal_reference(&cs).unwrap();
+        let computed_ref = test_message.proposal_reference(&cs).await.unwrap();
         assert_eq!(computed_ref, expected_ref.to_vec());
     }
 }
--- a/mls-rs/src/group/framing.rs
+++ b/mls-rs/src/group/framing.rs
@@ -419,6 +419,23 @@ impl MlsMessage {
 
         kp.to_reference(cipher_suite).await.map(Some)
     }
+
+    /// If this is a plaintext proposal, return the proposal reference that can be matched e.g. with
+    /// [`StateUpdate::unused_proposals`](super::StateUpdate::unused_proposals).
+    #[cfg(feature = "by_ref_proposal")]
+    #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    pub async fn proposal_reference<C: CipherSuiteProvider>(
+        &self,
+        cipher_suite: &C,
+    ) -> Result<Vec<u8>, MlsError> {
+        matches!(&self.payload, MlsMessagePayload::Plain(ptx) if ptx.content.content_type() == ContentType::Proposal)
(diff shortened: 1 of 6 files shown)


mls-rs/src/group/framing.rs
```diff
@@ -419,6 +419,23 @@ impl MlsMessage {
...
+    /// [`StateUpdate::unused_proposals`](super::StateUpdate::unused_proposals).
+    #[cfg(feature = "by_ref_proposal")]
+    #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    pub async fn proposal_reference<C: CipherSuiteProvider>(
+        &self,
+        cipher_suite: &C,
+    ) -> Result<Vec<u8>, MlsError> {
+        matches!(&self.payload, MlsMessagePayload::Plain(ptx) if ptx.content.content_type() == ContentType::Proposal)
+            .then_some(())
+            .ok_or(MlsError::UnexpectedMessageType)?;
```
stefunctional, line 433: Other methods return `None` or an empty `Vec` when the message has a different type. Should we be consistent here?

mls-rs/src/group/framing.rs
```diff
@@ -419,6 +419,23 @@ impl MlsMessage {
...
+
+        ProposalRef::from_bytes(cipher_suite, &self.payload.mls_encode_to_vec()?)
```
stefunctional, line 435: Shouldn't the bytes be the result of tls-encoding `AuthenticatedContent::from(ptx)` where `ptx` is the `PublicMessage`? As it stands, it looks like the computed ref would be different if the membership tag is `Some(_)`.
```rust
pub(crate) enum MlsMessagePayload {
    Plain(PublicMessage) = 1u16,
    // ...
}

pub(crate) struct PublicMessage {
    pub content: FramedContent,
    pub auth: FramedContentAuthData,
    pub membership_tag: Option<MembershipTag>,
}

pub struct AuthenticatedContent {
    pub(crate) wire_format: WireFormat,
    pub(crate) content: FramedContent,
    pub(crate) auth: FramedContentAuthData,
}
```

mulmarta: It's the same as the test confirms. These have the exact same byte format in this case and cloning everything into `AuthenticatedContent` seems suboptimal?

mls-rs/src/group/framing.rs
mulmarta: I wanted to avoid `Result<Option<Vec<u8>>>` which is complicated. But I don't feel strongly...

mls-rs/src/group/framing.rs
stefunctional: Your test sets the `membership_tag` to `None`, which makes the field completely disappear from the TLS-encoded output. If set to `Some(_)`, the TLS serialization between a `MlsMessagePayload` and its corresponding `AuthenticatedContent` would be different.

stefunctional (approved):

stefunctional (approved):

tomleavy (approved):

PR #62, opened 2024-01-10
Remove unused dependencies

mgeisler:
Hi,

I hope you'll like this small PR to remove some unused dependencies. I focused on mls-rs and its dependencies.

### Description of changes

This removes the unused enum-iterator, base64, num_enum, indexmap, and sha1 dependencies. The enum-iterator was mentioned in the code (but not used), the others were not mentioned at all outside of the Cargo.toml files.

This commit also makes spin, portable-atomic, and portable-atomic-util optional since they’re already used that way in the code.

Using

    cargo tree -p mls-rs -e normal --prefix none | sort -u | wc -l

to count unique transitive dependencies of mls-rs, this reduces the number from 60 to just 44.

### Testing

I made sure the code compiles and passes tests before and after:

```
% cargo check --all-targets
% cargo check --all-targets --all-features
% cargo test
% cargo test --all-features
```

### DIFF
--- a/mls-rs-identity-x509/Cargo.toml
+++ b/mls-rs-identity-x509/Cargo.toml
@@ -13,10 +13,13 @@ default = ["std"]
 std = ["mls-rs-core/std", "dep:thiserror"]
 
 [dependencies]
-async-trait = "0.1.74"
 mls-rs-core = { path = "../mls-rs-core", default-features = false, features = ["x509"], version = "0.17.0" }
 maybe-async = "0.2.7"
 thiserror = { version = "1.0.40", optional = true }
+
+# Async mode dependencies
+[target.'cfg(mls_build_async)'.dependencies]
+async-trait = "0.1.74"
 
 [dev-dependencies]
 mockall = "0.11"
--- a/mls-rs/Cargo.toml
+++ b/mls-rs/Cargo.toml
@@ -55,12 +55,11 @@ zeroize = { version = "1", default-features = false, features = ["alloc", "zeroi
 mls-rs-codec = { version = "0.5.0", path = "../mls-rs-codec", default-features = false}
 thiserror = { version = "1.0.40", optional = true }
 itertools = { version = "0.12.0", default-features = false, features = ["use_alloc"]}
-enum-iterator = "1.1.3"
 cfg-if = "1"
 debug_tree = { version = "0.4.0", optional = true }
-spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin_mutex", "portable_atomic"]}
-portable-atomic = { version = "1.5.1", default-features = false, features = ["critical-section"]}
-portable-atomic-util = { version = "0.1.2", default-features = false, features = ["alloc"]}
+spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin_mutex", "portable_atomic"] }
+portable-atomic = { version = "1.5.1", default-features = false, features = ["critical-section"] }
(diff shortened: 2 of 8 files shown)


mls-rs-identity-x509/Cargo.toml
```diff
@@ -13,7 +13,6 @@ default = ["std"]
 std = ["mls-rs-core/std", "dep:thiserror"]
 
 [dependencies]
-async-trait = "0.1.74"
```
stefunctional, line 16: I think code generated by `maybe-async` needs this. But maybe it can be moved to a `[target.'cfg(mls_build_async)'.dependencies]` section.

mgeisler: Thanks, I'll take a look at the build failures and see if I can fix them somehow!

stefunctional: Thank you for your contribution and taking the time to clean up these dependencies! Don't hesitate if you have questions.

mgeisler: Yes, your idea seems to work!

mls-rs/Cargo.toml
```diff
@@ -55,12 +55,11 @@ zeroize = { version = "1", default-features = false, features = ["alloc", "zeroi
...
 itertools = { version = "0.12.0", default-features = false, features = ["use_alloc"]}
-enum-iterator = "1.1.3"
 cfg-if = "1"
 debug_tree = { version = "0.4.0", optional = true }
-spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin_mutex", "portable_atomic"]}
-portable-atomic = { version = "1.5.1", default-features = false, features = ["critical-section"]}
-portable-atomic-util = { version = "0.1.2", default-features = false, features = ["alloc"]}
+spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin_mutex", "portable_atomic"], optional = true}
+portable-atomic = { version = "1.5.1", default-features = false, features = ["critical-section"], optional = true}
+portable-atomic-util = { version = "0.1.2", default-features = false, features = ["alloc"], optional = true}
```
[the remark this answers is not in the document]
mgeisler: Right... and from what I understand and [can read](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), you're doing things the recommended way by having a `std` feature instead of a `no_std` one.

I wonder what the usual work-around is in this case. The portable-atomic and portable-atomic-util crates are the ones I'm trying to avoid since I don't have them available in AOSP where I work.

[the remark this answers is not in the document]
mgeisler: Let me drop this part for now: I can patch things locally as needed and I think you're correct that there is no current good way of doing what I want.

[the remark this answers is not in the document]
stefunctional: I think we can't rely on features (at least as we've defined them) to do that. Other options are:
1. Adding these dependencies based on the platform -- this may be quite tedious and incomplete.
2. Adding another cfg like `mls_build_async` to enable/disable these dependencies.

I can check (2) with Tom and get back to you. This can be done in a separate PR if that helps.

stefunctional (approved):

tomleavy:
Hey @mgeisler, this looks great,  thanks for the contribution!

tomleavy (approved):

mgeisler:
> Hey @mgeisler, this looks great, thanks for the contribution!

You're very welcome!

PR #65, opened 2024-01-16
Depend on portable-atomic-* only if the target has no OS

stefunctional:
### Issues:

N/A

### Description of changes:

This makes the `portable-atomic*` dependencies conditional on the target not having an OS.

### Call-outs:

N/A

### Testing:

Built


tomleavy (approved):

mulmarta (approved):

stefunctional:
@mgeisler, may you check if the latest `main` resolves your issue regarding the `portable-atomic*` dependencies?

stefunctional:
Hi @mgeisler,

Right, `portable-atomic*` would only be used on targets that don't have an OS. It's not clear to me what that means for WASM.

There's a more precise `cfg` to test for atomic support in `core` at compile-time, so I opened #66  for that.

PR #68, opened 2024-01-23
Fix conditional compilation of proposal references

stefunctional:
Some uses were incorrectly guarded by the `external_client` feature.

### DIFF
--- a/mls-rs/src/group/message_processor.rs
+++ b/mls-rs/src/group/message_processor.rs
@@ -62,7 +62,7 @@ use super::proposal::CustomProposal;
 #[cfg(feature = "private_message")]
 use crate::group::framing::PrivateMessage;
 
-#[cfg(all(feature = "by_ref_proposal", feature = "external_client"))]
+#[cfg(feature = "by_ref_proposal")]
 use mls_rs_codec::{MlsDecode, MlsEncode, MlsSize};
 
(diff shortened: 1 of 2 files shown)


tomleavy (dismissed):

mls-rs/src/group/message_processor.rs
```diff
@@ -284,19 +284,19 @@ pub struct ProposalMessageDescription {
     pub proposal: Proposal,
     /// Plaintext authenticated data in the received MLS packet.
     pub authenticated_data: Vec<u8>,
-    #[cfg(all(feature = "by_ref_proposal", feature = "external_client"))]
+    #[cfg(feature = "by_ref_proposal")]
     pub(crate) proposal_ref: ProposalRef,
 }
```
mulmarta, line 289: Not related to this PR, but should this struct be non-exhaustive? Should this field be public like the other fields?

mulmarta (approved):

mls-rs/src/group/message_processor.rs
stefunctional: Good point. The cfg guard is also redundant. I pushed a commit to address all that. Thank you!

mulmarta (approved):

tomleavy (approved):

PR #72, opened 2024-01-29
Allow validating messages

mulmarta:
### Description of changes:

`GroupInfo`, `KeyPackage` and `Welcome` can now be validated using `process_message`, the same way as `MlsCiphertext` was validated by `ExternalGroup`. 

### Callout
The CI fails because of the aws-lc-sys problem on main.

### DIFF
--- a/mls-rs/src/external_client/group.rs
+++ b/mls-rs/src/external_client/group.rs
@@ -24,13 +24,13 @@ use crate::{
         snapshot::RawGroupState,
         state::GroupState,
         transcript_hash::InterimTranscriptHash,
-        validate_group_info, ContentType, ExportedTree, GroupContext, Roster,
+        validate_group_info, ContentType, ExportedTree, GroupContext, GroupInfo, Roster, Welcome,
     },
     identity::SigningIdentity,
     protocol_version::ProtocolVersion,
     psk::AlwaysFoundPskStorage,
     tree_kem::{node::LeafIndex, path_secret::PathSecret, TreeKemPrivate},
-    CryptoProvider, MlsMessage,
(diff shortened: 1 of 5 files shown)


stefunctional (approved):

mls-rs/src/external_client/group.rs
```diff
@@ -1270,4 +1294,57 @@ mod tests {
...
+        let alice = test_group_with_one_commit(TEST_PROTOCOL_VERSION, TEST_CIPHER_SUITE).await;
+        let mut server = make_external_group(&alice).await;
+
+        let kp = test_key_package_message(TEST_PROTOCOL_VERSION, TEST_CIPHER_SUITE, "john").await;
+
+        let update = server.process_incoming_message(kp.clone()).unwrap();
+        let kp = kp.into_key_package().unwrap();
+
+        assert_matches!(update, ExternalReceivedMessage::KeyPackage(update_kp) if update_kp == kp);
+    }
```
tomleavy, line 1326: This is an interesting one because it would be useful mostly when I am trying to figure out if a KeyPackage will work with an existing group as part of an add proposal?

tomleavy (dismissed):

mls-rs/src/external_client/group.rs
mulmarta: Right, that only checks that cipher suite and version match. Which made me realize we need a package validation function on the `ExternalClient` that the server can use to validate uploaded packages, which is added now.

tomleavy: 👍

stefunctional (approved):

tomleavy (approved):

PR #74, opened 2024-01-31
Make secret tree work with arbitrary indices

mulmarta:
### Description of changes:

This is first step towards morphing SecretTree into a PPRF which is a secret tree where indices are bigints (PPRF is a crypto primitive that may be useful for some MLS extensions).

Changes: Make `SecretTree` generic over a `TreeIndex` trait that does tree math. Also use a map instead of a vec since `SecretTree` is usually sparse. `TreeKemPublic` still only works for `u32` but should be possible to generalize.

### Testing:

All interop tests pass. There is a `SecretTree` test with `u64` indices.

### DIFF
--- a/mls-rs/src/group/secret_tree.rs
+++ b/mls-rs/src/group/secret_tree.rs
@@ -101,6 +101,8 @@ impl<T: TreeIndex> TreeSecretsVec<T> {
     fn set_node(&mut self, index: T, value: SecretTreeNode) {
         if let Some(i) = self.find_node(&index) {
             self.inner[i] = (index, value)
+        } else {
+            self.inner.push((index, value))
         }
     }
 
--- a/mls-rs/src/group/secret_tree.rs
+++ b/mls-rs/src/group/secret_tree.rs
@@ -110,10 +110,10 @@ pub struct SecretTree<T: MlsCodec> {
 }
 
 impl<T: TreeIndex + MlsCodec> SecretTree<T> {
-    pub(crate) fn empty(zero_leaf_count: T) -> SecretTree<T> {
+    pub(crate) fn empty() -> SecretTree<T> {
         SecretTree {
             known_secrets: Default::default(),
-            leaf_count: zero_leaf_count,
+            leaf_count: T::zero(),
         }
     }
 }
--- a/mls-rs/src/group/secret_tree.rs
+++ b/mls-rs/src/group/secret_tree.rs
@@ -101,7 +101,7 @@ pub struct SecretTree<T: TreeIndex> {
     leaf_count: T,
 }
 
-impl<T: TreeIndex + TreeIndex> SecretTree<T> {
+impl<T: TreeIndex> SecretTree<T> {
     pub(crate) fn empty() -> SecretTree<T> {
         SecretTree {
             known_secrets: Default::default(),
--- a/mls-rs/src/group/secret_tree.rs
+++ b/mls-rs/src/group/secret_tree.rs
@@ -18,9 +18,6 @@ use mls_rs_core::error::IntoAnyError;
 #[cfg(feature = "std")]
 use std::collections::HashMap;
 
-#[cfg(not(feature = "std"))]
-use alloc::collections::BTreeMap;
-
 use super::key_schedule::kdf_expand_with_label;
 
 pub(crate) const MAX_RATCHET_BACK_HISTORY: u32 = 1024;
@@ -82,9 +79,10 @@ struct TreeSecretsVec<T: TreeIndex> {
     #[cfg(feature = "std")]
     inner: HashMap<T, SecretTreeNode>,
     #[cfg(not(feature = "std"))]
-    inner: BTreeMap<T, SecretTreeNode>,
+    inner: Vec<(T, SecretTreeNode)>,
 }
 
+#[cfg(feature = "std")]
 impl<T: TreeIndex> TreeSecretsVec<T> {
     fn set_node(&mut self, index: T, value: SecretTreeNode) {
         self.inner.insert(index, value);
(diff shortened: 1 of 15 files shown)


mls-rs/src/tree_kem/math.rs
```diff
@@ -3,121 +3,132 @@
...
+macro_rules! impl_tree_stdint {
+    ($t:ty) => {
+        impl TreeIndex for $t {
+            fn root(&self) -> $t {
+                *self - 1
+            }
 
-    while x != root(n) {
-        let p = parent(x);
+            /// Panicks if `x` is even.
```
stefunctional, line 80:
```suggestion
            /// Panicks if `x` is even in debug, overflows in release.
```

mls-rs/src/tree_kem/math.rs
```diff
@@ -3,121 +3,132 @@
...
+            fn left_unchecked(&self) -> Self {
+                *self ^ (0x01 << (level(*self) - 1))
+            }
 
-        d.push(if x < p {
-            right_unchecked(p)
-        } else {
-            left_unchecked(p)
-        });
+            /// Panicks if `x` is even.
```
stefunctional, line 85:
```suggestion
            /// Panicks if `x` is even in debug, overflows in release.
```

mls-rs/src/tree_kem/math.rs
```diff
@@ -3,121 +3,132 @@
...
-    }
+            fn zero() -> Self {
+                0
+            }
+        }
 
-    Ok(d)
+        fn level(x: $t) -> u32 {
+            x.trailing_ones()
+        }
```
stefunctional, line 122: [Nit] This will be problematic (duplicate name) if we ever want to invoke this macro twice in the same module. It's fine for now though.

mls-rs/src/tree_kem/math.rs
```diff
@@ -3,121 +3,132 @@
...
-
-use crate::client::MlsError;
+use core::fmt::Debug;
 
 use super::node::LeafIndex;
 
-pub fn level(x: u32) -> u32 {
-    x.trailing_ones()
-}
+pub trait TreeIndex: Sized + Send + Sync + Eq + Clone + Debug {
```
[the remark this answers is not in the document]
mulmarta: 1. I was hoping to get rid of the other `TreeIndex` in a subsequent PR... Would that make sense? (It's anyway better type safety than having `TreeIndex`, `LeafIndex` and `u32` behave the same)
2. `SecretTree` uses `MlsCodec` only but I agree that one trait is simpler. Removed `MlsCodec`.

mls-rs/src/tree_kem/math.rs
```diff
@@ -3,121 +3,132 @@
...
-    if x > 2 * n - 1 {
-        return Err(MlsError::InvalidTreeIndex);
+        path
     }
+}
 
-    let mut d = Vec::new();
-    let mut m = 1 << (level(x) + 1);
+#[derive(Clone, PartialEq, Eq, core::fmt::Debug)]
+pub struct CopathNode<T: Clone + PartialEq + Eq + core::fmt::Debug> {
```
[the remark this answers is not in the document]
mulmarta: I see, good to know about deriving. Thank you!

mls-rs/src/tree_kem/math.rs
mulmarta: Yeah, that's why there's a `mod` around `impl_tree_stdint!(u64)`. Removed now

mls-rs/src/group/secret_tree.rs
```diff
@@ -104,12 +96,12 @@ impl<T: MlsCodec> TreeSecretsVec<T> {
...
 
 #[derive(Clone, Debug, PartialEq, MlsEncode, MlsDecode, MlsSize)]
-pub struct SecretTree<T: MlsCodec> {
+pub struct SecretTree<T: TreeIndex> {
     known_secrets: TreeSecretsVec<T>,
     leaf_count: T,
 }
 
-impl<T: TreeIndex + MlsCodec> SecretTree<T> {
+impl<T: TreeIndex + TreeIndex> SecretTree<T> {
```
stefunctional, line 104:
```suggestion
impl<T: TreeIndex> SecretTree<T> {
```

stefunctional (dismissed):

stefunctional (approved):

mls-rs/src/group/secret_tree.rs
```diff
@@ -83,54 +77,35 @@ impl From<Zeroizing<Vec<u8>>> for TreeSecret {
...
-impl DerefMut for TreeSecretsVec {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.0
-    }
+#[derive(Clone, Debug, PartialEq, MlsEncode, MlsDecode, MlsSize, Default)]
+struct TreeSecretsVec<T: TreeIndex> {
+    #[cfg(feature = "std")]
+    inner: HashMap<T, SecretTreeNode>,
+    #[cfg(not(feature = "std"))]
+    inner: BTreeMap<T, SecretTreeNode>,
```
tomleavy, line 85: I don't think this important right now, but I believe this would be the only place BTreeMap is used in the minimal configuration and thus would drive up binary size. Maybe this could be `Vec<T, SecretTreeNode>` instead if binary size becomes an issue?

tomleavy (dismissed):

mls-rs/src/group/secret_tree.rs
mulmarta: Fixed @tomleavy

tomleavy (approved):

stefunctional (approved):

PR #76, opened 2024-02-05
Add a serde feature and derive serde for most structs used by providers

mulmarta:
Allows more convenient providers implementations, e.g. with JSONs for debug. Changes mechanical.


mls-rs-core/Cargo.toml
```diff
@@ -14,15 +14,16 @@ exclude = ["test_data"]
 default = ["std", "rfc_compliant", "fast_serialize"]
 arbitrary = ["std", "dep:arbitrary"]
 fast_serialize = ["mls-rs-codec/preallocate"]
-std = ["mls-rs-codec/std", "zeroize/std", "safer-ffi-gen?/std", "dep:thiserror"]
+std = ["mls-rs-codec/std", "zeroize/std", "safer-ffi-gen?/std", "dep:thiserror", "serde?/std"]
 rfc_compliant = ["x509"]
 ffi = ["dep:safer-ffi", "dep:safer-ffi-gen"]
 x509 = []
 test_suite = ["dep:serde", "dep:serde_json", "dep:hex", "dep:itertools"]
+serde = ["dep:serde"]
```
stefunctional, line 22:
```suggestion
serde = ["dep:serde", "zeroize/serde"]
```
So that `serde` is not pulled unconditionally by `zeroize`.

stefunctional (approved):

mls-rs-core/src/crypto.rs
```diff
@@ -68,6 +70,7 @@ impl AsRef<[u8]> for HpkePublicKey {
     all(feature = "ffi", not(test)),
     safer_ffi_gen::ffi_type(clone, opaque)
 )]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct HpkeSecretKey(#[mls_codec(with = "mls_rs_codec::byte_vec")] Vec<u8>);
```
tomleavy, line 74: I think before we had a serde with here that would base64 the byte arrays, otherwise you get really ugly JSON with each byte as a number

tomleavy (commented):
I think we just need to put a base64 transform anywhere that we have serde where we use bytevec for the mls side

mls-rs-core/src/crypto.rs
mulmarta: I thought about it but didn't know how to integrate with `cfg_attr` without causing a mess. Is there a way?

tomleavy: Should be, let me take a look on Monday

mls-rs-core/src/lib.rs
```diff
@@ -24,3 +24,50 @@ pub use mls_rs_codec;
...
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+    pub fn serialize<S: Serializer>(v: &Vec<u8>, s: S) -> Result<S::Ok, S::Error> {
+        Vec::<u8>::serialize(&v, s)
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(d: D) -> Result<Vec<u8>, D::Error> {
+        Vec::<u8>::deserialize(d)
+    }
+}
```
tomleavy, line 73: The cleaner way to do this is to just have one vec_serde that checks the value of [is_human_readable](https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable) and adjusts accordingly without the feature flag

mls-rs-core/src/lib.rs
```diff
@@ -24,3 +24,42 @@ pub use mls_rs_codec;
...
+
+#[cfg(feature = "serde")]
+pub mod vec_serde {
+    use alloc::{string::String, vec::Vec};
+    use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+
+    pub fn serialize<S: Serializer>(v: &Vec<u8>, s: S) -> Result<S::Ok, S::Error> {
+        if s.is_human_readable() {
+            String::serialize(&hex::encode(v), s)
```
stefunctional, line 52: `hex` already does that for you:
```suggestion
            hex::serde::serialize(v, s)
```

mls-rs-core/src/lib.rs
```diff
@@ -24,3 +24,42 @@ pub use mls_rs_codec;
...
+        } else {
+            Vec::<u8>::serialize(&v, s)
```
stefunctional: This is the recommended way to serialize bytes with serde (it can be more efficient and works better with formats that have a custom representation for bytes):
```suggestion
            s.serialize_bytes(&v)
```

mls-rs-core/src/lib.rs
```diff
@@ -24,3 +24,42 @@ pub use mls_rs_codec;
...
+        }
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(d: D) -> Result<Vec<u8>, D::Error> {
+        if d.is_human_readable() {
+            hex::decode(String::deserialize(d)?).map_err(Error::custom)
```
stefunctional, line 60:
```suggestion
            hex::serde::deserialize(d)
```

mls-rs-core/src/lib.rs
```diff
@@ -24,3 +24,42 @@ pub use mls_rs_codec;
...
+        } else {
+            Vec::<u8>::deserialize(d)
```
stefunctional, line 62: This should use the `serde_bytes` crate for efficient and flexible serialization of bytes and do this:
```rust
serde_bytes::deserialize(d)
```

`serde_bytes` is a common dependency when using `serde`, and it does not bring any additional dependency.

mls-rs/Cargo.toml
```diff
@@ -38,7 +38,8 @@ std = ["mls-rs-core/std", "mls-rs-codec/std", "mls-rs-identity-x509?/std", "hex/
 
 ffi = ["dep:safer-ffi", "dep:safer-ffi-gen", "mls-rs-core/ffi"]
 
-serde = ["mls-rs-core/serde", "zeroize/serde", "dep:serde"]
+serde = ["mls-rs-core/serde", "zeroize/serde", "dep:serde", "dep:hex"]
```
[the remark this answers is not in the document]
stefunctional: N/A anymore now that the `hex_serde` feature was removed.

stefunctional (approved):

tomleavy (approved):

PR #80, opened 2024-02-21
Async build

mulmarta:
Add macros to make the SQLite build in the async mode. This way it can be used e.g. with async identity or crypto providers. Build WebCrypto provider only for wasm arch.


stefunctional (approved):
Thank you for taking care of the redundant imports which I also faced when I pushed my branch last night!

mrdowns (approved):

Issue #81, opened 2024-02-21
[uniffi] Expose a UniFFI-compatible API


mgeisler:
### Description of feature:

I would like to use mls-rs from languages like Swift and Kotlin — and [UniFFI](https://mozilla.github.io/uniffi-rs/) seems like a great tool for this. It auto-generates high-level Rust bindings in a number of languages.

### Use case:

Swift and Kotlin bindings will make it easy to create iOS and Android apps which use mls-rs.

### Implementation discussion (Optional)

I'm planning on creating a small crate which makes the necessary adaptations of the mls-rs API to make it compatible with UniFFI.

PR #82, opened 2024-02-21
Introduce mls-rs-uniffi

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

This new create is a thin wrapper around mls-rs which exposes a subset of it’s interface using a [UniFFI-compatible][1] API.

We will expand this over the next few weeks to be a full API usable by messaging apps on different platforms.

[1]: https://mozilla.github.io/uniffi-rs/

### Call-outs:

The code currently only works with `mls_build_async`, but this should be fixable in the future. From talking to @mulmarta, there should be a way to use traits to handle both cases.

### Testing:

I've mostly looked at the auto-generated Kotlin code to see that it has a useful shape. I expect us to add/remove things from the API in the next few weeks.


mulmarta:
To add more context to the sync build problems, the following code shows the issue and one not too satisfying solution. @stefunctional maybe you have ideas how to improve?

```
uniffi::setup_scaffolding!();

#[maybe_async::must_be_sync]
async fn bar() {}

#[derive(uniffi::Object)]
struct Foo;

/*
// This does not work because maybe_async must be called _before_ uniffi
// so that uniffi sees the sync function. On the other hand, uniffi must be called
// on the impl block which is higher up (and maybe_async can't be called on
// the impl block).

#[uniffi::export]
impl Foo {
    #[maybe_async::must_be_sync]
    async fn foo(&self) {
        bar().await
    }
}
*/

// This works
#[maybe_async::must_be_sync]
trait FooType {
    async fn foo(&self);
}

#[maybe_async::must_be_sync]
#[uniffi::export]
impl FooType for Foo {
    async fn foo(&self) {
        bar().await
    }
}
```

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+
+/// Extract the basic credential identifier from a  from a key package.
+fn signing_identity_to_identifier(
+    signing_identity: &mls_rs_core::identity::SigningIdentity,
+) -> Result<Vec<u8>, Error> {
+    match &signing_identity.credential {
+        identity::Credential::Basic(credential) => Ok(credential.identifier.clone()),
+        _ => Err(MlsError::RequiredCredentialNotFound(BasicCredential::credential_type()).into()),
+    }
+}
```
mulmarta, line 383: Suggestion: the same functionality is achieved by `return BasicIdentityProvider::new().identity(&signing_identity)` with `use mls_rs_core::identity::IdentityProvider`

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+        Ok(mls_message.into())
+    }
+
+    /// Propose and commit the removal of a member.
+    ///
+    /// The member is representated by the key package in `member`.
+    ///
+    /// See [`mls_rs::group::CommitBuilder::remove_member`] for details.
+    pub async fn remove_member(&self, member: Arc<SigningIdentity>) -> Result<CommitOutput, Error> {
+        let identifier = signing_identity_to_identifier(&member.inner)?;
```
mulmarta, line 436: Can it be `member: &SigningIdentity`?

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+    /// Propose to remove a member from this group.
+    ///
+    /// The member is representated by the key package in `member`.
+    /// The result is the welcome message to send to this member.
+    ///
+    /// See [`mls_rs::group::Group::propose_remove`] for details.
+    pub async fn propose_remove_member(
+        &self,
+        key_package_message: Arc<Message>,
+    ) -> Result<Message, Error> {
```
mulmarta, line 456: I think this should take `signing_idantity` not `key_package` to match `remove_member`?

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+        let message = arc_unwrap_or_clone(message);
+        let mut group = self.inner().await;
+        match group.process_incoming_message(message.inner).await? {
+            group::ReceivedMessage::ApplicationMessage(application_message) => {
+                let sender = Arc::new(
+                    self.index_to_identity(application_message.sender_index)
+                        .await?
+                        .into(),
+                );
+                let data = application_message.authenticated_data;
```
mulmarta, line 492: This should be `let data = application_message.data().to_vec()`. The `authenticated_data` will usually be empty -- this is sent _unencrypted_ along with the message

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+                    }
+                    _ => todo!("External and NewMember proposal senders are not supported"),
+                };
+                let proposal = Arc::new(proposal_message.proposal.into());
+                Ok(ReceivedMessage::Proposal { sender, proposal })
+            }
+            // TODO: group::ReceivedMessage::GroupInfo does not have any
+            // public methods (unless the "ffi" Cargo feature is set).
+            // So perhaps we don't need it?
+            group::ReceivedMessage::GroupInfo(_) => Ok(ReceivedMessage::GroupInfo),
```
mulmarta, line 516: It has methods in [here](https://github.com/awslabs/mls-rs/blob/main/mls-rs/src/group/group_info.rs#L29) but the type GroupInfo is private due to a forgotten export which ought to be fixed.

stefunctional:
fMeow/maybe-async-rs#24 should help with the interoperability between `#[uniffi::export]` and `maybe_async` macros.

mgeisler:
> [fMeow/maybe-async-rs#24](https://github.com/fMeow/maybe-async-rs/pull/24) should help with the interoperability between `#[uniffi::export]` and `maybe_async` macros.

Thanks @stefunctional, that sounds great! I'll take a look!

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,524 @@
...
+    /// The user is identified by `id`, which will be used to create a
+    /// basic credential together with the signature keypair.
+    ///
+    /// See [`mls_rs::Client::builder`] for details.
+    #[uniffi::constructor]
+    pub fn new(
+        id: Vec<u8>,
+        signature_keypair: SignatureKeypair,
+        cipher_suite: CipherSuite,
+    ) -> Self {
```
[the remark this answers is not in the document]
mgeisler: Great idea, makes a lot of sense!

mls-rs-uniffi/src/lib.rs
mgeisler: Yeah, I agree!

mls-rs-uniffi/src/lib.rs
mgeisler: Yes, much better!

If/when `Group::propose_add` and `ClientBuilder::add_member` would stop taking ownership of the key package passed to them, then we can also simplify the signatures for `add_member` and `propose_add_member`.

mls-rs-uniffi/src/lib.rs
mgeisler: Thanks! I need to get some unit tests going for this ASAP...

mls-rs-uniffi/src/lib.rs
mgeisler: I thought the methods there were only meant to be used with safer-ffi?

I'll just leave this open for now, then we can return to it when we have more of the API built out.

mgeisler:
> > [fMeow/maybe-async-rs#24](https://github.com/fMeow/maybe-async-rs/pull/24) should help with the interoperability between `#[uniffi::export]` and `maybe_async` macros.
> 
> Thanks @stefunctional, that sounds great! I'll take a look!

This worked beautifully, I can now compile both with and without `mls_build_async`!

mls-rs-uniffi/src/lib.rs
mgeisler: I hope the TODO can do for now — we'll find this quickly when we start building on top of the API here.

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+    }
+
+    /// Commit the addition of a member.
+    ///
+    /// The member is representated by the key package in `member`.
+    /// The result is the welcome message to send to this member.
+    ///
+    /// See [`mls_rs::group::CommitBuilder::add_member`] for details.
+    pub async fn add_member(&self, key_package: Arc<Message>) -> Result<CommitOutput, Error> {
+        let key_package = arc_unwrap_or_clone(key_package);
```
mulmarta, line 419: Can be later, but it would be nice to make this `add_members(&self, key_packages: Arc<Vec<Message>>)` so that I can add many members in one commit.

mulmarta (approved):

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+#[uniffi::export]
+impl Client {
+    /// Create a new client.
+    ///
+    /// The user is identified by `id`, which will be used to create a
+    /// basic credential together with the signature keypair.
+    ///
+    /// See [`mls_rs::Client::builder`] for details.
+    #[uniffi::constructor]
+    pub fn new(id: Vec<u8>, signature_keypair: SignatureKeypair) -> Self {
```
tomleavy, line 260: Lets make an issue to make this a SigningIdentity and expose support for x509 as well. We can probably do that after the provider traits are exposed

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+#[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+async fn signing_identity_to_identifier(
+    signing_identity: &identity::SigningIdentity,
+) -> Result<Vec<u8>, Error> {
+    let identifier = basic::BasicIdentityProvider::new()
+        .identity(signing_identity, &mls_rs::ExtensionList::new())
+        .await
+        .map_err(|err| err.into_any_error())?;
+    Ok(identifier)
+}
```
tomleavy, line 394: I think when we have identity provider exposed this won't be necessary anymore since we can just store the identity provider in the group struct and call `self.identity_provider.identity`

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+#[uniffi::export]
+impl Group {
+    /// Perform a commit of received proposals (or an empty commit).
+    ///
+    /// TODO: ensure `path_required` is always set in
+    /// [`MlsRules::commit_options`](`mls_rs::MlsRules::commit_options`).
+    ///
+    /// Returns the resulting commit message. See
+    /// [`mls_rs::Group::commit`] for details.
+    pub async fn commit(&self) -> Result<Message, Error> {
```
tomleavy, line 406: We probably want this to return CommitOutput as there are other properties in there you will require. Can you just add the function to get the commit message from CommitOutput for now?

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+        Ok(commit_output.commit_message.into())
+    }
+
+    /// Commit the addition of a member.
+    ///
+    /// The member is representated by the key package in `member`.
+    /// The result is the welcome message to send to this member.
+    ///
+    /// See [`mls_rs::group::CommitBuilder::add_member`] for details.
+    pub async fn add_member(&self, key_package: Arc<Message>) -> Result<CommitOutput, Error> {
```
tomleavy, line 418: For performance it is better that this is add_members and just takes a vec as input. Since we don't have the commit_builder exposed probably all of these functions should take a vec so that you can run multiple operations

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+            group,
+            group_info_extensions,
+        })
+    }
+}
+
+#[derive(Clone, Debug, uniffi::Object)]
+pub struct CommitOutput {
+    _inner: mls_rs::group::CommitOutput,
+}
```
mulmarta, line 333: There seems to be no way to extract commit and welcome messages from this struct?

mulmarta:
Another thing @mgeisler there seems to be no way to `apply_pending_commit`? Should `process_incoming_message` maybe take message or ack?

mls-rs-uniffi/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+    ) -> Result<ReceivedMessage, Error> {
+        let message = arc_unwrap_or_clone(message);
+        let mut group = self.inner().await;
+        match group.process_incoming_message(message.inner).await? {
+            group::ReceivedMessage::ApplicationMessage(application_message) => {
+                let sender = Arc::new(
+                    self.index_to_identity(application_message.sender_index)
+                        .await?
+                        .into(),
+                );
```
mulmarta, line 495: `self.index_to_identity` will try to lock the already locked `self.inner()` and hang forever. Fix

```
       let message = {
            let mut group = self.inner().await;
            group.process_incoming_message(message.inner).await?
        };

        match message {
            group::ReceivedMessage::ApplicationMessage(application_message) => {
```

mls-rs-uniffi/src/lib.rs
mgeisler: Sure, let me change this! I think the type has to be `Vec<Arc<Message>>` since `Vec<Message>` isn't compatible with UniFFI.

mls-rs-uniffi/src/lib.rs
mgeisler: Sounds good, I created #85 for this.

mls-rs-uniffi/src/lib.rs
mgeisler: Nice! I'm still learning how all these types fit together, but your idea here is clearly much nicer.

mls-rs-uniffi/src/lib.rs
mgeisler: No, it's not hooked up to anything yet — several of them are not used yet. That should definitely change soon!

mls-rs-uniffi/src/lib.rs
mgeisler: Let me start by returning the full `CommitOutput` here like I do elsewhere. I also added a few methods to `CommitOutput` now like @mulmarta suggested below.

mls-rs-uniffi/src/lib.rs
mgeisler: Yep, I've changes the signatures to take/return vectors now.

mls-rs-uniffi/src/lib.rs
mgeisler: This is now fixed as per @tomleavy's comment above.

mls-rs-uniffi/src/lib.rs
mgeisler: Thanks for testing this! I think I've found another way of fixing it: turning `index_to_identity` into a free function which takes a `&mls_rs::Group` as an argument. 

That should avoid the double-locking and it might suggest a pattern we should follow: `self.inner()` calls in the methods since this is prone to dead-locks as you point out.

mulmarta (approved):

tomleavy (approved):

PR #86, opened 2024-02-25
[uniffi] Add support for custom GroupStateStorage interface

tomleavy:
### Issues:
#81

### Description of changes:

* Add a new FFI safe version of GroupStateStorage that will erase generics from the interface.
* Create a Config struct that can take an `Arc<dyn GroupStateStorage>` as input that works with uniffi and converts it into a GroupStateStorageWrapper that we can then pass to mls-rs upon configuration.

### Call-outs:

I am unsure if this actually works in practice yet. The code does generate for Swift, but it looks rather odd (which might just be a quirk of how uniffi handles traits more generally.

### Testing:

Keeping this as a draft until someone can create a swift / kotlin / python program that proves the basic strategy is ok.


tomleavy:
@mulmarta I think this helps us not go too crazy in the short term. If we can get this to work then we can decide later if mls-rs should erase these generics or not and the change won't break this interface since they are already erased.

mulmarta:
This will only build in async once
* UniFFI updates -- support of async trait [was added a week ago](https://github.com/mozilla/uniffi-rs/pull/1981)
* async-trait is added to async dependencies of mls-rs-uniffi

mgeisler:
I tried using the code here from Python via #87. I got the error @mulmarta talked about:

```
Traceback (most recent call last):
  File "/tmp/run-python-rj1JKa/script.py", line 17, in <module>
    client = Client(b"alice", signature_keypair, client_config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/run-python-rj1JKa/mls-rs-uniffi-5cb1d76f88c59931/mls_rs_uniffi.py", line 1236, in __init__
    _UniffiConverterTypeClientConfig.lower(client_config))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/run-python-rj1JKa/mls-rs-uniffi-5cb1d76f88c59931/mls_rs_uniffi.py", line 428, in lower
    cls.write(value, builder)
  File "/tmp/run-python-rj1JKa/mls-rs-uniffi-5cb1d76f88c59931/mls_rs_uniffi.py", line 2475, in write
    _UniffiConverterTypeGroupStateStorage.write(value.group_state_storage, buf)
  File "/tmp/run-python-rj1JKa/mls-rs-uniffi-5cb1d76f88c59931/mls_rs_uniffi.py", line 2091, in write
    buf.write_u64(cls.lower(value))
                  ^^^^^^^^^^^^^^^^
  File "/tmp/run-python-rj1JKa/mls-rs-uniffi-5cb1d76f88c59931/mls_rs_uniffi.py", line 2080, in lower
    return value._uniffi_clone_pointer()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/run-python-rj1JKa/mls-rs-uniffi-5cb1d76f88c59931/mls_rs_uniffi.py", line 1990, in _uniffi_clone_pointer
    return _rust_call(_UniffiLib.uniffi_mls_rs_uniffi_fn_clone_groupstatestorage, self._pointer)
                                                                                  ^^^^^^^^^^^^^
AttributeError: 'PythonGroupStateStorage' object has no attribute '_pointer'
test sync_tests::test_client_new ... FAILED
```

I looked some more and the support for foreign traits (implementing a Rust trait in another language) seems very new:

- https://github.com/mozilla/uniffi-rs/pull/1818
- https://github.com/mozilla/uniffi-rs/pull/1902
- https://github.com/mozilla/uniffi-rs/pull/1969
- https://github.com/mozilla/uniffi-rs/pull/1972

(and possibly more).

The second of these made me aware of a `#[uniffi::export(with_foreign)]` syntax: the `with_foreign` attribute corresponds to [this short section in the unreleased manual](https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/udl/interfaces.md#foreign-implementations).

It's not clear to me what the overall state of all this is, but if I use `#[uniffi::export(with_foreign)]` on `pub trait GroupStateStorage`, I'm forced to change the `&[u8]` types to `Vec<u8>` (seems like a good change, since it indicates that the type takes ownership, which you would expect when passed in from Python).

When generate the Python bindings, `GroupStateStorage` now looks like this:

```python
class GroupStateStorage(typing.Protocol):

    def state(self, group_id: "bytes"):
        raise NotImplementedError

    def epoch(self, group_id: "bytes", epoch_id: "int"):
        raise NotImplementedError

    def write(self, state: "GroupState",
              epoch_inserts: "typing.List[EpochRecord]",
              epoch_updates: "typing.List[EpochRecord]"):
        raise NotImplementedError

    def max_epoch_id(self, group_id: "bytes"):
        raise NotImplementedError
```

before it looked like this:

```python
class GroupStateStorage:

    _pointer: ctypes.c_void_p

    def __del__(self):
        # In case of partial initialization of instances.
        pointer = getattr(self, "_pointer", None)
        if pointer is not None:
            _rust_call(_UniffiLib.uniffi_mls_rs_uniffi_fn_free_groupstatestorage, pointer)

    def _uniffi_clone_pointer(self):
        return _rust_call(_UniffiLib.uniffi_mls_rs_uniffi_fn_clone_groupstatestorage, self._pointer)
```

I can now generate a `Client` using an empty Python implementation of the storage trait:

```python
from mls_rs_uniffi import *

signature_keypair = generate_signature_keypair(CipherSuite.CURVE25519_AES128)

class PythonGroupStateStorage(GroupStateStorage):
    def state(self, group_id: "bytes"):
        raise NotImplementedError
    def epoch(self, group_id: "bytes",epoch_id: "int"):
        raise NotImplementedError
    def write(self, state: "GroupState",epoch_inserts: "typing.List[EpochRecord]",epoch_updates: "typing.List[EpochRecord]"):
        raise NotImplementedError
    def max_epoch_id(self, group_id: "bytes"):
        raise NotImplementedError

group_state_storage = PythonGroupStateStorage()
client_config = ClientConfig(group_state_storage)
client = Client(b"alice", signature_keypair, client_config)
```

I pushed the change here: https://github.com/mgeisler/mls-rs/commit/35b19d5f4df3a07d7b75c69dff29e8309b98b007.

tomleavy:
Hey @mgeisler funny I was deep into looking through issues and found the with_foreign stuff as well. Made the edits and now I think we are good!

mulmarta (approved):

mls-rs-uniffi/src/config/group_state.rs
```diff
@@ -0,0 +1,123 @@
...
+        epoch_inserts: Vec<ET>,
+        epoch_updates: Vec<ET>,
+    ) -> Result<(), Self::Error>
+    where
+        ST: mls_rs_core::group::GroupState + MlsEncode + MlsDecode + Send + Sync,
+        ET: mls_rs_core::group::EpochRecord + MlsEncode + MlsDecode + Send + Sync,
+    {
+        let state = Arc::new(GroupState {
+            id: state.id(),
+            data: state.mls_encode_to_vec().unwrap(),
```
stefunctional, line 104: Should an error be returned instead?

mls-rs-uniffi/src/config/group_state.rs
```diff
@@ -0,0 +1,123 @@
...
+        });
+
+        let epoch_to_record = |v: ET| {
+            Arc::new(EpochRecord {
+                id: v.id(),
+                data: v.mls_encode_to_vec().unwrap(),
```
stefunctional, line 110: Should an error be returned instead?

stefunctional (approved):

stefunctional (approved):

mulmarta:
> Hey @mgeisler funny I was deep into looking through issues on uniffi and found the `with_foreign` stuff as well. Made the edits and now I think we are good!
> 
> I also added async_trait in there, so if you patch in mainline uniffi it should work (I think)

The need to patch in mainline uniffi also breaks the CI. You can ignore the test for now https://github.com/awslabs/mls-rs/blob/main/mls-rs-uniffi/src/lib.rs#L594

tomleavy:
@mulmarta removed the file / ignored the test for now to avoid rewriting it. Lets just wait until uniffi updates and rewrite the test based on whatever progress we make on the API in the meantime

stefunctional (approved):

mls-rs-uniffi/src/config.rs
```diff
@@ -0,0 +1,43 @@
...
+use mls_rs_crypto_openssl::OpensslCryptoProvider;
+
+use self::group_state::GroupStateStorageWrapper;
+
+mod group_state;
+
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+#[uniffi(flat_error)]
+#[non_exhaustive]
+pub enum FFICallbackError {
```
mgeisler, line 17: Did you consider just extending the `Error` enum instead to keep the API smaller?

mulmarta (approved):

mls-rs-uniffi/src/config.rs
tomleavy: We can do that as a follow up, was keeping this isolated and didn't merge them together.

PR #87, opened 2024-02-25
Execute UniFFI bindings via a simple Python test

mgeisler:
### Issues:

Addresses #81

### Description of changes:

This uses the UniFFI test helpers to compile and make the `libmls_rs_uniffi.so` file available to the Python script.

The test is very simple here: it simply executes the Python script, which then has to return with a zero exit code. If it fails, the output is printed and you then have to debug the Python code by hand.

There is no facility to pass data back to the Rust unit test — it would be great to have this, but right now it's a one-way communication only.

### Call-outs:

I would have liked to use `$CARGO_TARGET_TMPDIR` to place the temporary files in a stable place under `target/`. However, this variable is only set for integration tests, not unit tests. So the script now ends up somewhere in your system temp directory.

We could decide to only use integration tests for these tests, let me know what you think! We could also write a bit more code to find a good place under `target/` without relying on the environment variable.

### Testing:

Use `cargo test`!

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -17,6 +17,9 @@
 //!
 //! [UniFFI]: https://mozilla.github.io/uniffi-rs/
 
+#[cfg(test)]
+pub mod test_utils;
+
 use std::sync::Arc;
 
 #[cfg(not(mls_build_async))]
@@ -570,5 +573,22 @@ impl Group {
                 Ok(ReceivedMessage::KeyPackage { key_package })
             }
         }
+    }
+}
+
+#[cfg(all(test, not(mls_build_async)))]
+mod sync_tests {
+    use crate::test_utils::run_python;
+
+    #[test]
+    fn test_generate_signature_keypair() -> Result<(), Box<dyn std::error::Error>> {
+        run_python(
+            r#"
+from mls_rs_uniffi import CipherSuite, generate_signature_keypair
+
+signature_keypair = generate_signature_keypair(CipherSuite.CURVE25519_AES128)
+assert signature_keypair.cipher_suite == CipherSuite.CURVE25519_AES128
+"#,
+        )
     }
 }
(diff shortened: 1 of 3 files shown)


mgeisler:
@mulmarta, I think this more or less implements your idea of testing using Python. You're right: it's super easy to get going with just a few lines of code! This also very quickly shows that we need a lot more fields or getters on the types to actually make them usable from any of the high-level target langauges :smile: 

@tomleavy, this should also help with testing #86.

mgeisler:
> ## [Codecov](https://app.codecov.io/gh/awslabs/mls-rs/pull/87?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs) Report
> All modified and coverable lines are covered by tests ✅

It would be amazing to have this integrate with the coverage reports — I think that would require making the Python code call into the coverage-instrumented `.so` file somehow.

mls-rs-uniffi/src/lib.rs
```diff
@@ -17,6 +17,9 @@
 //!
 //! [UniFFI]: https://mozilla.github.io/uniffi-rs/
 
+#[cfg(test)]
```
mgeisler, line 20: Ah, I see how that you use a pattern of having a `test_util` Cargo feature for this. I can update to use the same pattern here.

mulmarta: I think what you did is simpler? AFAIK we only used the feature when a `test_util` module was used in different crates. E.g. we have "test_util" feature for HPKE so that different crypto provider crates using the HPKE crate can execute known-answer HPKE tests

mgeisler: > I think what you did is simpler?

I guess: I had not thought about this being a matter of cross-crate dependencies. I'm happy to leave it like it is now if you think it's fine.

tomleavy (approved):

mulmarta (approved):

mls-rs-uniffi/src/lib.rs
mulmarta: Thanks!

PR #88, opened 2024-02-26
Make Snapshot public

mulmarta:
Ideal world: `GroupStateStorage::write(&mut self, state: &Snapshot)` which matches `KeyPackageStorage::insert(&mut self, id: Vec<u8>, pkg: KeyPackageData)`. Access to specific secrets can be convenient.

Reality: moving `Snapshot` to `mls-rs-core` is terrible. To alleviate this, this PR makes Snapshot public so that storage implementor can deserialize it. We should still consider passing bytes instead of the `StateData` trait.


mls-rs/src/group/snapshot.rs
```diff
@@ -23,83 +23,81 @@ use super::proposal_cache::{CachedProposal, ProposalCache};
...
-
-        #[cfg(not(feature = "tree_index"))]
-        let public_tree = {
-            let mut tree = TreeKemPublic::new();
-            tree.nodes = state.public_tree.nodes.clone();
-            tree
-        };
+impl RawGroupState<'_> {
+    pub(crate) fn export(mut state: GroupState) -> RawGroupState<'static> {
+        // TODO this clone isn't necessary if we make the storage trait take bytes
```
stefunctional, line 79: This comment does not seem relevant to the code below?

mls-rs/src/group/snapshot.rs
```diff
@@ -23,83 +23,81 @@ use super::proposal_cache::{CachedProposal, ProposalCache};
...
-    pub(crate) public_tree: TreeKemPublic,
-    pub(crate) interim_transcript_hash: InterimTranscriptHash,
-    pub(crate) pending_reinit: Option<ReInitProposal>,
-    pub(crate) confirmation_tag: ConfirmationTag,
+    pub proposals: Vec<(ProposalRef, CachedProposal)>,
+    pub public_tree: ExportedTree<'a>,
+    pub interim_transcript_hash: InterimTranscriptHash,
+    pub pending_reinit: Option<ReInitProposal>,
+    pub confirmation_tag: ConfirmationTag,
+    tree_internals: TreeKemPublic,
```
stefunctional, line 74: This field ought to have a comment stating that its `nodes` field is likely empty and useless (except in one specific function where it gets repopulated).

PR #89, opened 2024-02-26
Add more Python FFI examples

mulmarta:


tomleavy:
@mulmarta is there a way to just have a .py file and import it as a string using include_str! ... That would help be able to at least use a normal editing flow which is helpful

mulmarta:
> @mulmarta is there a way to just have a .py file and import it as a string using include_str! ... That would help be able to at least use a normal editing flow which is helpful

Sure but due to Github problems it broke the PR. See #90

PR #90, opened 2024-02-26
Add more Python FFI examples

mulmarta:


stefunctional (approved):

mgeisler:
Nice @mulmarta! I see you also started using `include_str!`... shortly after playing with a nono-trivial example, I found that it's super annoying to embed Python code in the raw string literal :smile:

I think it could be nice to change the argument to a path instead: the path could feed into the UniFFI testing function, which in turn uses it to generate a stable hash for the temporary directory for each test. We currently create files in `/tmp`, but we could improve this to use the `target` folder instead.

PR #91, opened 2024-02-27
[uniffi] Add a `generate_python_tests!` macro

mgeisler:
### Issues:

Addresses #81

### Description of changes:

Replace hand-written tests with a macro. The macro takes the name of a Python script inside the `test_bindings/` directory and uses this to construct a unit test.

This saves a bit of boiler-plate code at the point where the test is created. This comes at the expense of more “magic” in the form of the macro. The consistency of no boiler-plate code might make this worth the cost — but this is up for debate!


stefunctional (commented):
Have you considered moving the tests to be integration tests instead? This would solve the output directory dilemma as `env!("CARGO_TARGET_TMPDIR")` could be used. This is also more in line with what `UniFFITestHelper`, which is used by `run_script` under the hood, [prescribes](https://docs.rs/uniffi_testing/latest/uniffi_testing/struct.UniFFITestHelper.html):
> Used from a integration test (a .rs file in the tests/ directory)

tomleavy (dismissed):

mgeisler:
> Hey @mgeisler looks good but conflicts with the changes I just merged, take a look when you get a chance.

Will do! I wasn't sure how much you would like to lean into macros? Personally, I don't like them too much, but I also don't like boiler-plate code :smile: Now that I've made the macro generate both the sync and async test cases, I feel that it's becoming quite useful.

tomleavy:
> > Hey @mgeisler looks good but conflicts with the changes I just merged, take a look when you get a chance.
> 
> Will do! I wasn't sure how much you would like to lean into macros? Personally, I don't like them too much, but I also don't like boiler-plate code 😄 Now that I've made the macro generate both the sync and async test cases, I feel that it's becoming quite useful.

Maybe the suggestion @stefunctional has above would help reduce boilerplate and macros?

mgeisler:
> Have you considered moving the tests to be integration tests instead? This would solve the output directory dilemma as `env!("CARGO_TARGET_TMPDIR")` could be used. This is also more in line with what `UniFFITestHelper`, which is used by `run_script` under the hood, [prescribes](https://docs.rs/uniffi_testing/latest/uniffi_testing/struct.UniFFITestHelper.html):
> 
> > Used from a integration test (a .rs file in the tests/ directory)

Yes, that's a great point — I agree that it would be better aligned with the expectations of the helper.

I normally prefer in-line unit tests... that's actually why I let `run_python` take a string literal at first since it invites you to just write the Python code directly where you need it. However, that idea is flawed: as @mulmarta quickly found out, you'll want an external file with proper syntax highlighting when you start writing non-trivial tests. Those "unit tests" don't feel very much like a unit test at that point :smile: 

Given this, I think you're right that we should just use integration tests instead. Let me rewrite the PR to do that.

mulmarta (approved):

mgeisler:
I rewrote things to use a single integration test called `scenarios.rs` instead. This can conveniently house the macro and be expanded with more and more tests as we add more Python code around this.

stefunctional (approved):

mulmarta (approved):

PR #96, opened 2024-03-04
Tell `rustfmt` about the edition we use

mgeisler:
My editor (Emacs) likes to format individual files by sending them to `rustfmt`. This fails without a configuration file to set the edition since `rustfmt` (when executed alone, not via `cargo fmt`) defaults to the good old Rust 2015 edition.


mgeisler:
Related to this, have you considered using

```toml
imports_granularity = "Module"
```

to tell `rustfmt` to turn things like

```rust
use mls_rs_core::{
    crypto::{CipherSuite, HpkePublicKey, HpkeSecretKey},
    error::IntoAnyError,
};
```

into

```rust
use mls_rs_core::crypto::{CipherSuite, HpkePublicKey, HpkeSecretKey};
use mls_rs_core::error::IntoAnyError;
```

I find the module-scoped imports easier to read since they are more self-contained (less nesting).

However, the `rustfmt` option is unstable, so it only works on nightly — `rustfmt` from stable Rust will ignore it with a warning. Crucially, stable `rustfmt` will not revert the formatting done by nightly `rustfmt` since the default is to not touch the imports.

tomleavy (approved):

PR #100, opened 2024-03-05
Added import and export functions to the GroupStateStorage api

### LINKED ISSUE #99: Introduce the ability to import/export group state from an existing client.
### Description of feature:

We would like to be able to import/export group state from/to an existing client.

### Use case:

This can be useful when a client is first initialising a group and wants to pre populate that group with a member that does not yet exist. In this scenario, the client would generate key packages and private group state for the to be created member and then would export the data so that it can be securely transferred to and imported by the new member once it comes online.

There are two ways this functionality could be implemented.

1. By exposing the snapshot.rs module to pub visibility. This makes some types visible globally and may result in some pollution of the public API for what is admittedly a niche case.
2. By introducing export and import methods in the storage class. This avoids the main drawback listed above but comes at the cost of potential data corruption on import as the storage API has no reliable way to validate the arbitrary bytes being imported. This could be something that we avoid by providing our own impl of "GroupState + MlsEncode + MlsDecode + Send + Sync" to the write function but at that point we are just recreating the Snapshot api...

CaioSym:
### Issues:

Resolves #99 

### Description of changes:

Currently, there is no way to public import/export private group state from the mls crate as a whole. This PR adds import/export functionality to the GroupStateStorage trait and its implementations.

### Call-outs:

After implementing this. I'm more and more convinced that perhaps option 1 in #99 would be a better approach. The current impl forces every new impl of GroupStateStorage to implement and test import/export logic which is duplicative and the inability to verify the validity of the imported bytes is particularly dangerous imo. Please let me know if you would like me to pivot to option 1.

### Testing:

Unit tests.

### DIFF
--- a/mls-rs-core/src/group/group_state.rs
+++ b/mls-rs-core/src/group/group_state.rs
@@ -82,4 +82,16 @@ pub trait GroupStateStorage: Send + Sync {
     /// The [`EpochRecord::id`] value that is associated with a stored
     /// prior epoch for a particular group.
     async fn max_epoch_id(&self, group_id: &[u8]) -> Result<Option<u64>, Self::Error>;
+
+
+    async fn export_snapshot_data(
+        &self,
+        group_id: &[u8],
+    ) -> Result<Option<Vec<u8>>, Self::Error>;
+
+    async fn import_snapshot_data(
+        &self,
+        group_id: &[u8],
+        group_snapshot: Vec<u8>,
+    ) -> Result<(), Self::Error>;
 }
--- a/mls-rs-provider-sqlite/src/group_state.rs
+++ b/mls-rs-provider-sqlite/src/group_state.rs
@@ -264,6 +264,24 @@ impl GroupStateStorage for SqLiteGroupStateStorage {
             .transpose()
             .map_err(|e| SqLiteDataStorageError::DataConversionError(e.into()))
     }
+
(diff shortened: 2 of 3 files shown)


tomleavy:
Hey @CaioSym , we had some other pending changes in this same area that I think actually simplify things for you. Check out #94 . Either way I would prefer that it not be required for every storage provider to implement this. Probably best to just implement the function on SQLite storage / InMemory storage directly or do something like `pub trait ExportableGroupStateStorage: GroupStateStorage` and add the additional methods there

mulmarta:
Hey @CaioSym thanks for looking into this!

I think that with #94 import is subsumed by `write`? For export, I agree with @tomleavy that not _every_ storage has to implement it. For your use case you can use a specific implementation of storage like in this example [main.rs](https://github.com/awslabs/mls-rs/files/14503245/main.rs.txt). Is the trait actually necessary?

Unfortunately, making `Snapshot` and its internals public would require exposing a _lot_ of internal API. I was thinking one compromise would be to make `Snapshot` public but with all fields private, and then add public functions on `Snapshot` that allow accessing selected parts that may be needed? Could this address your use case?

CaioSym:
Closing this as no longer needed as its covered by #94. Will instead work on just exposing the storage to avoid the current issue with having to clone it.

PR #101, opened 2024-03-05
[uniffi] Rename enum variants to avoid collision

### LINKED ISSUE #98: [uniffi] Enum variants cannot match existing types
### Problem:

Enum variants cannot have names that collide with other types: UniFFI will generate incorrect Kotlin code in that case. See https://github.com/mozilla/uniffi-rs/issues/1853.

### Solution:

We should rename our enum variants as needed until this is fixed upstream. It affects us in two cases at the moment:

* https://github.com/awslabs/mls-rs/blob/de125ef942564d3b4029eb0e4c3ab16656b1955d/mls-rs-uniffi/src/lib.rs#L178-L182
* https://github.com/awslabs/mls-rs/blob/de125ef942564d3b4029eb0e4c3ab16656b1955d/mls-rs-uniffi/src/lib.rs#L189-L190

In both cases, the generated code will recursively refer back to the enum variant instead of the top-level type.

We should ideally fix this upstream, but we'll probably just rename the variants for now.

mgeisler:
Because of mozilla/uniffi-rs#1853, our enum variants must have names that are different from all top-level types.

I simply added a `_`, but other ideas are of course very welcome.

### Issues:

Resolves #98

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -175,8 +175,10 @@ pub enum ReceivedMessage {
     /// A new commit was processed creating a new group state.
     Commit { committer: Arc<SigningIdentity> },
 
+    // TODO(mgeisler): rename to `Proposal` when
+    // https://github.com/awslabs/mls-rs/issues/98 is fixed.
     /// A proposal was received.
-    Proposal {
+    ReceivedProposal {
         sender: Arc<SigningIdentity>,
         proposal: Arc<Proposal>,
     },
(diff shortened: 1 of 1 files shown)


tomleavy:
What does the _ generate in the target language?

mulmarta (approved):

stefunctional (dismissed):

mgeisler:
> It turns out the trailing `_` is stripped, at least for Kotlin.

This in turn is a hint that I need to setup CI for the generated code in more languages than just Python. Since we care about Kotlin, I'll try setting that up at some point.

stefunctional (approved):

tomleavy (approved):

tomleavy:
Thanks @mgeisler

PR #108, opened 2024-03-08
[Bugfix] Derive external pub extension from the new and not old key schedule

mulmarta:
The added test fails on main with `InvalidConfirmationTag`. The reason is generating external join info using old epoch's key schedule.

### DIFF
--- a/.github/workflows/no_std_build.yml
+++ b/.github/workflows/no_std_build.yml
@@ -7,16 +7,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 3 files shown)


stefunctional (approved):

stefunctional:
I'm confused. It seems the test fails but only on wasm?

mulmarta:
> I'm confused. It seems the test fails but only on wasm?

Yeah, the cipher suite it used wasn't supported by wasm crypto provider. Fixed.

stefunctional:
> Yeah, the cipher suite it used wasn't supported by wasm crypto provider. Fixed.

Thank you for explaining!

tomleavy (approved):

PR #110, opened 2024-03-11
Add APIs to client to expose the internal storages so that they may be consulted directly.

### LINKED ISSUE #99: Introduce the ability to import/export group state from an existing client.

CaioSym:
### Issues:

Resolves #99 

### Description of changes:

#94 made the API for storage simply expose Vec<u8>s which is perfect for importing and exporting data. Unfortunately, there is no current way to access the underlying storage of the group other than setting a storage directly on the config builder, which requires a use of `clone` and most importantly, that the group storage being set have some kind of internal mutability. This PR adds APIs to the Client which expose the storages.

### Call-outs:

While only access to group state storage is needed for this feature, it seems prudent to expose the other storages in case they are needed in the future. I'm happy to revert those changes if we prefer however.

### Testing:

The change simply exposes some internal values, It does not therefore add any tests.

### DIFF
--- a/mls-rs/src/client.rs
+++ b/mls-rs/src/client.rs
@@ -707,6 +707,27 @@ where
     pub fn key_package_extensions(&self) -> ExtensionList {
         self.config.key_package_extensions()
     }
+
+    /// The [KeyPackageStorage](crate::KeyPackageStorage) that
+    /// this client was configured to use.
+    #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
+    pub fn key_package_store(&self) -> <C as ClientConfig>::KeyPackageRepository {
+        self.config.key_package_repo()
+    }
+
+    /// The [PreSharedKeyStorage](crate::PreSharedKeyStorage) that
+    /// this client was configured to use.
+    #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
+    pub fn secret_store(&self) -> <C as ClientConfig>::PskStore {
+        self.config.secret_store()
+    }
+
+    /// The [GroupStateStorage](crate::GroupStateStorage) that
+    /// this client was configured to use.
+    #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
+    pub fn group_state_storage(&self) -> <C as ClientConfig>::GroupStateStorage {
+        self.config.group_state_storage()
+    }
 }
 
 #[cfg(test)]


tomleavy (approved):

mls-rs/src/client.rs
```diff
@@ -707,6 +707,27 @@ where
     pub fn key_package_extensions(&self) -> ExtensionList {
         self.config.key_package_extensions()
     }
+
+    /// The [KeyPackageStorage](crate::KeyPackageStorage) that
+    /// this client was configured to use.
+    #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
+    pub fn key_package_store(&self) -> <C as ClientConfig>::KeyPackageRepository {
```
mgeisler, line 714: Just from reading the signature, I'm surprised this isn't returned as a borrowed type? That is, `&<C as ClientConfig>::KeyPackageRepository`.

Is it perhaps expected that you `impl KeyPackageStorage for `&CustomStorage` when you have a custom storage class `CustomStorage`?

CaioSym: These functions simply copy the signature from the methods in Config.

The underlying Config API defines the types of all storages as their respective storage trait + Clone which means any access to config for the storage is already doing a move under the hood.

 I'll be honest I'm not sure why Config itself does that vs just exposing a reference... Maybe due to some use-cases that require mutability?

tomleavy: Is there any reason why this function in particular can't return a borrow even though internally we clone when the client creates a group?

CaioSym: Yes. I can't just change this signature to return a reference because config itself does the clone. This means the functions I'm introducing themselves own the storage clones returned by config, and those clones would go out of scope when the function returns unless they are returned by the function.

Trying to simply return a reference from these functions will result in:
```
cannot return reference to temporary value
returns a reference to data owned by the current functionrustc
client.rs(722, 10): temporary value created here
```

in order to do that I'd need to change config itself to return references... which has much broader implications.

mgeisler: > Yes. I can't just change this signature to return a reference because config itself does the clone.

Ah, I see... I think you are talking about code like this in `ConfigInner`:

https://github.com/awslabs/mls-rs/blob/2d92710155695c1cc07b9ca952e261bbc882c4be/mls-rs/src/client_builder.rs#L702-L716

That does look a bit odd to me — it looks like there is an underlying assumption that it's cheap to clone all these types and that the types refer back to the same underlying storage afterwards.



> The design here was that we have one storage for all groups (which is convenient for apps). So, this storage is `clone`d from `Client` to each group instance, whenever the client joins or creates a group.

Would it be possible for `Client` to own the storage, but put it in an `Arc` internally. Something like

```rust
    #[derive(Clone, Debug)]
    pub struct ConfigInner<Kpr, Ps, Gss, Ip, Pr, Cp> {
        pub(crate) settings: Settings,
        pub(crate) key_package_repo: Arc<Kpr>,
        pub(crate) psk_store: Arc<Ps>,
        pub(crate) group_state_storage: Arc<Gss>,
```

instead of

https://github.com/awslabs/mls-rs/blob/2d92710155695c1cc07b9ca952e261bbc882c4be/mls-rs/src/client_builder.rs#L920-L924

When creating/joining a `Group`, these fields would the be cloned (but since they're `Arc`s, the clone is cheap).

Without knowing much about how all this fits together, I feel this models the ownership better between `Client` and `Group`.

tomleavy: @stefunctional ^^

stefunctional: As @mulmarta mentioned, the different stores are shared between `Client` and `Group`s and we didn't want to tie the lifetime of a group to that of the client used to instantiate it.

`ClientConfig` methods return associated types assumed to be cheaply clonable. This is more versatile than forcing all stores to be `Arc`-ed, as not all store implementations may need to be. E.g. a storage implementation that would communicate with some storage service through HTTP could use `reqwest::Client` internally and not need to be wrapped in `Arc`. It's a common pattern in Rust to encapsulate how shared ownership is achieved (e.g. `reqwest::Client`, `tokio::sync::mpsc::Sender`); this allows using `Arc` internally or something else if it can be more efficient. This also prevents layers of `Arc`s when the inner type already uses `Arc`.

Regarding returning a reference to the storage, it's not possible with the current design as all `Group` knows is that its `config` implements `ClientConfig` and the trait methods return owned values. Given that cloning a store should most likely end up incrementing a ref count or duplicating a file descriptor, that should be cheap as long as it's not done in a tight loop, in which case getting the storage outside the loop should help.

CaioSym: So just to sanity check. Are we fine with these APIs being added "as is"?

mulmarta (approved):

stefunctional (approved):

mls-rs/src/client.rs
mgeisler: Thanks for explaining, I was not familiar with this pattern! It feels like to me like the `Clone` bound is meant to capture a few things:

- it's probably cheap to clone the storage
- the storage is probably still connected to each other afterwards

The "probably" above is what feels odd to me :smile: Using an `Arc` directly would codify this more strongly, but see what you say about layers of `Arc`s.

PR #111, opened 2024-03-12
Add `client_config_default` function for easy client configuration

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

The goal of this PR is to simply the steps needed to setup a simple client. It adds a new function which returns a default client config with an in-memory group state storage backend.

The new `GroupStateStorageAdapter` struct allows us to use any mls-rs group state storage, so we could easily surface the Sqlite storage as well now.

Since I was working with the storage trait, I also took the opportunity to flatten the errors a little. Right now, we don't seem to have a use case for separate errors, so I feel this makes things a little simpler overall.

### Testing:

New Python integration test.

### DIFF
--- a/mls-rs-uniffi/src/config.rs
+++ b/mls-rs-uniffi/src/config.rs
@@ -1,12 +1,14 @@
-use std::{fmt::Debug, sync::Arc};
+use std::fmt::Debug;
+use std::sync::Arc;
 
 use mls_rs::{
     client_builder::{self, WithGroupStateStorage},
     identity::basic,
+    storage_provider::in_memory::InMemoryGroupStateStorage,
 };
 use mls_rs_crypto_openssl::OpensslCryptoProvider;
 
-use self::group_state::GroupStateStorage;
+use self::group_state::{GroupStateStorage, GroupStateStorageAdapter};
 use crate::Error;
 
 pub mod group_state;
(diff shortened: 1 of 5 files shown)


mls-rs-uniffi/src/config.rs
```diff
@@ -1,43 +1,79 @@
...
 #[derive(Debug, Clone, uniffi::Record)]
 pub struct ClientConfig {
     pub group_state_storage: Arc<dyn GroupStateStorage>,
 }
+
+// TODO(mgeisler): turn into an associated function when UniFFI
+// supports them: https://github.com/mozilla/uniffi-rs/issues/1074.
+/// Create a client config with an in-memory group state storage.
+#[uniffi::export]
+pub fn client_config_default() -> ClientConfig {
```
mgeisler, line 73: Maybe I should add that I'm not attached to the name here: it could also be something like `with_in_memory_storage` if you like.

I would also like to have an easy way to instantiate the Sqlite storage, similar to [`ClientBuilder::new_sqlite`](https://docs.rs/mls-rs/latest/mls_rs/client_builder/struct.ClientBuilder.html#method.new_sqlite). Perhaps it could be even simpler and just take a path name to the database and then configure all three storages to use a Sqlite database under that path.

mulmarta (approved):

mls-rs-uniffi/src/config.rs
mulmarta: I like the SQLite idea, we could make them provide a path and a key -- let's make it encrypted by default?

mgeisler: That sounds nice indeed — I haven't looked at the Sqlite implementation yet, but does it support on-disk encryption directly? Or do you suggest that we encrypt the file on disk somehow?

tomleavy (approved):

PR #112, opened 2024-03-12
[uniffi] Add more methods to `CommitOutput`

mgeisler:
### Issues:

Addresses #81

### Description of changes:

This adds `ratchet_tree` and `group_info` methods to `CommitOutput`.

### Testing:

None yet, I will add a test after merging #111 since that PR will make it trivial again to setup a new client.

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -433,6 +433,48 @@ impl Client {
 }
 
 #[derive(Clone, Debug, uniffi::Object)]
+pub struct RatchetTree {
+    inner: mls_rs::group::ExportedTree<'static>,
+}
+
+impl From<mls_rs::group::ExportedTree<'static>> for RatchetTree {
+    fn from(inner: mls_rs::group::ExportedTree<'static>) -> Self {
+        Self { inner }
+    }
+}
+
+#[uniffi::export]
+impl RatchetTree {
+    /// Encode the ratchet tree in MLS encoding.
+    pub fn to_bytes(&self) -> Result<Vec<u8>, Error> {
+        self.inner.to_bytes().map_err(Into::into)
+    }
+
+    /// Return size of ratchet tree in MLS encoding.
+    pub fn byte_size(&self) -> u64 {
+        self.inner.byte_size().try_into().unwrap()
+    }
+}
+
+impl RatchetTree {
+    // TODO(mgeisler): merge with #[uniffi::export] impl above when
+    // https://github.com/mozilla/uniffi-rs/issues/1074 is fixed.
+    /// Decode a ratched tree from its MLS encoding.
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, Error> {
+        let exported_tree = mls_rs::group::ExportedTree::from_bytes(bytes)?;
+        Ok(exported_tree.into())
+    }
+}
+
+// TODO(mgeisler): remove this when associated functions are supported
+// by UniFFI: https://github.com/mozilla/uniffi-rs/issues/1074.
+#[uniffi::export]
+/// Decode a ratched tree from its MLS encoding.
+pub fn ratchet_tree_from_bytes(bytes: &[u8]) -> Result<RatchetTree, Error> {
+    RatchetTree::from_bytes(bytes)
+}
+
+#[derive(Clone, Debug, uniffi::Object)]
 pub struct CommitOutput {
     inner: mls_rs::group::CommitOutput,
 }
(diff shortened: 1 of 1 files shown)


mls-rs-uniffi/src/lib.rs
```diff
@@ -420,6 +420,51 @@ impl Client {
...
+        self.inner.to_bytes().map_err(Into::into)
+    }
+
+    /// Return size of ratchet tree in MLS encoding.
+    pub fn byte_size(&self) -> u64 {
+        #[cfg(not(target_pointer_width = "64"))]
+        compile_error!("mls-rs-uniffi expects a 64-bit pointer width");
+
+        self.inner.byte_size().try_into().unwrap()
+    }
```
mulmarta, line 447: Out of curiosity, what does it have to do with pointer width? Can't we convert whatever `usize` is to `u64` (any tree size will fit there for sure)?

mgeisler: Maybe I wrote something silly here... My intention is to do a lossless conversion of the `usize` to `u64`. However, `TryFrom` will fail if we're on a 32-bit machine and so I tried to give a better compiler error in that case.

I would really have liked to just write `self.inner.byte_size() as u64`,but the `as` conversions are frowned upon these days. On a a 64-bit machine, the [`try_into()` call ends up calling `as`](https://doc.rust-lang.org/stable/src/core/convert/num.rs.html#390).

mulmarta: Why would it fail on a 32-bit machine, though? We can convert `u32` to `u64`, no?

tomleavy:
@mgeisler just do you know it seems that uniffi does some not so nice things with memory management in Kotlin so unless you make things like CommitOutput a Record the users of this library will be required to "close" all handles. In my experiments I switched to just marshaling everything that doesn't have internal state.

mls-rs-uniffi/src/lib.rs
stefunctional: AFAIK, on existing platforms, this conversion never fails. If there ever are architectures with 128-bit pointers, such a conversion would become fallible and apparently that's why the `std` lib does not provide `From<usize> for u64`.

mgeisler: > Why would it fail on a 32-bit machine, though? We can convert `u32` to `u64`, no?

Heh, that's true... I guess I was thinking of it backwards :smile: 

> AFAIK, on existing platforms, this conversion never fails. If there ever are architectures with 128-bit pointers, such a conversion would become fallible and apparently that's why the `std` lib does not provide `From<usize> for u64`.

Yes, that matches my understanding as well.

I think I can simplify this by just removing the `compile_error!` call then.

mgeisler:
> @mgeisler just do you know it seems that uniffi does some not so nice things with memory management in Kotlin so unless you make things like CommitOutput a Record the users of this library will be required to "close" all handles. In my experiments I switched to just marshaling everything that doesn't have internal state.

Funny, we were just discussing this in a meeting here. I will be happy to follow this advice to make life easier for the developers who will end up using this.

mgeisler:
> unless you make things like CommitOutput a Record the users of this library will be required to "close" all handles.

To be more precise, I believe you're talking about the same thing as in this comment: https://github.com/mozilla/uniffi-rs/pull/1869#pullrequestreview-1868535066.

There I tried figuring out why the Kotlin code has a big warning about memory management. My understanding is that apps can see their memory usage stay artificially high if the system is under load and the cleaner thread (and GC) doesn't run. This in turn can result in an untimely death in a system like Android.

It's not clear to me how much of a problem this is now — as far as I understand, Kotlin has a `use` pattern which people can use to auto-close the objects.

In this case, I think we have two choices for `CommitOutput`:

- `CommitOutput` could be a UniFFI record. This means that we eagerly marshal the commit message, the welcome messages, the ratchet tree, and the group info.
- `CommitOutput` could be a UniFFI object. This is the path I started down on. We could still marshal the types returned by the various methods.

Marshalling eagerly sounds more expensive, but since commits are done rarely(?), I guess paying this cost can be acceptable.

tomleavy:
Quite honestly I think their concerns are misplaced and it's ok to deallocate native heap memory on the finalizer. Maybe good for you to look into with people who understand the Android JVM more, but SWIG for example does this for Java when wrapping C libraries and it is a non issue. 

My understanding is that the finalizer is not guaranteed but typically would only be skipped if the app is shutting down or something where the deallocation doesn't matter anyway, but I could be wrong. 

For now I chose record because yes it's expensive but actually the app will very likely call every getter anyway which might be more expensive overall. I'm basically using record whenever objects are returned from a function and not input to another function to start.

tomleavy (approved):

stefunctional (approved):

mgeisler:
> Quite honestly I think their concerns are misplaced and it's ok to deallocate native heap memory on the finalizer.

Yeah, I have the same understanding. However, I learnt last week that the UniFFI code uses a new background thread for the deallocations. I'm not sure why this is needed, but I can see how using a thread can affect the deallocations in case the thread is starved for resources.

We'll basically keep an eye on this as we get into testing an actual Hello World program in Kotlin. Then we can adjust as needed.

> For now I chose record because yes it's expensive but actually the app will very likely call every getter anyway which might be more expensive overall. I'm basically using record whenever objects are returned from a function and not input to another function to start.

Yes, I plan on following this approach as well going forward (and will convert the existing APIs into this form).

PR #116, opened 2024-03-15
[uniffi] Remove `KeyPackage` from the API

mgeisler:
### Issues:

Addresses #81

### Description of changes:

We don’t think people will need to access the key packages and their fields. As an example, when adding members to a group, you use messages, not key packages.

Related to #81, reverts most of #102. I left the `ProtocolVersion` behind, but I can remove it too if people think we don’t need it.

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -158,53 +158,6 @@ impl TryFrom<mls_rs::ProtocolVersion> for ProtocolVersion {
     }
 }
 
-/// Wrapper around a [`mls_rs::crypto::HpkePublicKey`].
-#[derive(Clone, Debug, uniffi::Record)]
-pub struct HpkePublicKey {
-    pub key: Vec<u8>,
-}
-
-impl From<mls_rs::crypto::HpkePublicKey> for HpkePublicKey {
-    fn from(key: mls_rs::crypto::HpkePublicKey) -> Self {
-        Self { key: key.into() }
-    }
-}
-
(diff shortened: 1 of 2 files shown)


mgeisler:
As discussed with @mulmarta earlier today.

mls-rs-uniffi/src/lib.rs
```diff
@@ -262,11 +205,8 @@ pub enum ReceivedMessage {
     GroupInfo,
     /// Validated welcome message.
     Welcome,
-
-    // TODO(mgeisler): rename to `KeyPackage` when
-    // https://github.com/awslabs/mls-rs/issues/98 is fixed.
     /// Validated key package.
-    ValidatedKeyPackage { key_package: Arc<KeyPackage> },
```
mulmarta, line 269: We might add `signing_identity` / `credential` here to allow verifying that we received key package for the client we actually wanted to add.

mulmarta (approved):

mls-rs-uniffi/src/lib.rs
mgeisler: Yes, would you like me to do this before we remove `KeyPackage` or can we do it down the road?

mulmarta: Just leaving a note, doesn't have to be now. Approved the PR.

mgeisler: Sounds good!

stefunctional (approved):

PR #123, opened 2024-03-18
Bump versions for 0.39 release

tomleavy:
Prep for releasing 0.39 crates

### DIFF
--- a/mls-rs-crypto-awslc/Cargo.toml
+++ b/mls-rs-crypto-awslc/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-awslc"
-version = "0.10.0"
+version = "0.11.0"
 edition = "2021"
 description = "AWS-LC based CryptoProvider for mls-rs"
 homepage = "https://github.com/awslabs/mls-rs"
@@ -9,20 +9,20 @@ keywords = ["mls", "mls-rs", "aws-lc"]
 license = "Apache-2.0 OR MIT"
 
 [dependencies]
-aws-lc-rs = "1.6.1"
+aws-lc-rs = "1.6.2"
 aws-lc-sys = { version = "0.13.0" }
-mls-rs-core = { path = "../mls-rs-core", version = "0.17.1" }
-mls-rs-crypto-hpke = { path = "../mls-rs-crypto-hpke", version = "0.8.0" }
-mls-rs-crypto-traits = { path = "../mls-rs-crypto-traits", version = "0.9.0" }
-mls-rs-identity-x509 = { path = "../mls-rs-identity-x509", version = "0.10.0" }
+mls-rs-core = { path = "../mls-rs-core", version = "0.18.0" }
+mls-rs-crypto-hpke = { path = "../mls-rs-crypto-hpke", version = "0.9.0" }
+mls-rs-crypto-traits = { path = "../mls-rs-crypto-traits", version = "0.10.0" }
+mls-rs-identity-x509 = { path = "../mls-rs-identity-x509", version = "0.11.0" }
 thiserror = "1.0.40"
 zeroize = { version = "1", features = ["zeroize_derive"] }
 maybe-async = "0.2.10"
 
 [dev-dependencies]
 assert_matches = "1.5.0"
-mls-rs-core = { path = "../mls-rs-core", version = "0.17.1", features = ["test_suite"] }
-mls-rs-crypto-hpke = { path = "../mls-rs-crypto-hpke", version = "0.8.0", features = ["test_utils"] }
+mls-rs-core = { path = "../mls-rs-core", version = "0.18.0", features = ["test_suite"] }
+mls-rs-crypto-hpke = { path = "../mls-rs-crypto-hpke", version = "0.9.0", features = ["test_utils"] }
 futures-test = "0.3.25"
 
 [target.'cfg(mls_build_async)'.dependencies]
--- a/mls-rs-crypto-hpke/Cargo.toml
+++ b/mls-rs-crypto-hpke/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-hpke"
-version = "0.8.0"
+version = "0.9.0"
 edition = "2021"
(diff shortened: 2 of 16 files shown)


mulmarta (approved):

mls-rs-crypto-awslc/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-awslc"
-version = "0.10.0"
+version = "0.10.1"
```
stefunctional, line 3:
```suggestion
version = "0.11.0"
```
This crate depends on `mls-rs-core` and that is a part of its public API (e.g. for the traits it defines), so bumping `mls-rs-core` to `0.18.0` is a breaking change for this crate.

mls-rs-crypto-hpke/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-hpke"
-version = "0.8.0"
+version = "0.8.1"
```
stefunctional, line 3:
```suggestion
version = "0.9.0"
```
Because `mls-rs-core` is part of the public API of this crate (e.g. with `HpkeContextS`)

mls-rs-crypto-openssl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-openssl"
-version = "0.8.0"
+version = "0.8.1"
```
stefunctional, line 3:
```suggestion
version = "0.9.0"
```

mls-rs-crypto-rustcrypto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-rustcrypto"
-version = "0.9.0"
+version = "0.9.1"
```
stefunctional, line 3:
```suggestion
version = "0.10.0"
```

mls-rs-crypto-traits/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-traits"
-version = "0.9.0"
+version = "0.9.1"
```
stefunctional, line 3:
```suggestion
version = "0.10.0"
```
Because of `mls-rs-core` that is part of the public API of this crate (e.g. `IntoAnyError`).

mls-rs-crypto-webcrypto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-webcrypto"
-version = "0.3.0"
+version = "0.3.1"
```
stefunctional, line 3:
```suggestion
version = "0.4.0"
```

mls-rs-identity-x509/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-identity-x509"
-version = "0.10.0"
+version = "0.10.1"
```
stefunctional, line 3:
```suggestion
version = "0.11.0"
```

stefunctional (commented):
Some of the version increments need to be adjusted. Any crate that has `mls-rs-core` as a dependency such that some of the definitions in `mls-rs-core` are part of the other crate's public API requires a minor bump.

`mls-rs-ffi` version was not bumped. Is it not going to be published?

mls-rs-crypto-traits/Cargo.toml
mgeisler: I think you're right if one wants to be conservative. However, a more fine-grained question would be if `mls-rs-crypto-traits` exports types with breaking changes?

In the case of `IntoAnyError`, I see that it was most recently modified [8 months ago](https://github.com/awslabs/mls-rs/blame/5235fc8cb15295498389f65a9ac6ba175cc392d9/mls-rs-core/src/error.rs#L38). There have been several releases since, so any breaking change in `IntoAnyError` should have been taken into account a long time ago, no?

stefunctional: The issue is that as far as rustc is concerned, `mls_rs_core::error::IntoAnyError` from `mls-rs-core 0.17.1` is always different from `mls_rs_core::error::IntoAnyError` from `mls-rs-core 0.18.0` (it does not matter if the definition of the trait hasn't changed). So if I have a project using `mls-rs-core 0.17.1` and `mls-rs-crypto-traits 0.9.0` and I run `cargo update`, my project now uses `mls-rs-core 0.17.1` (no change) and `mls-rs-crypto-traits 0.9.1` which was updated and now uses the definitions from `mls-rs-core 0.18.0` which are considered different from the ones from `mls-rs-core 0.17.1`, and so my code might break.

`IntoAnyError` is just an example of a public trait from `mls-rs-core` that is used in the public API of `mls-rs-crypto-traits`, but any type or trait from `mls-rs-core` used in the public API of `mls-rs-crypto-traits` can become an issue if the version is not updated appropriately. Let's simplify to highlight what the problem is.

Let's consider that `mls-rs-core` defines a type `Foo`, `mls-rs-crypto-traits` defines `fn take_foo(_: Foo)`, and my crate depends on both, instantiates a `Foo` and calls `take_foo`. After running `cargo update`, I get `mls-rs-crypto-traits 0.9.1` which depends on `mls-rs-core 0.18.0`, but my crate still depends on `mls-rs-core 0.17.1` directly (it wasn't updated because it's not considered semver-compatible). So now my crate instantiates `Foo` from `mls-rs-core 0.17.1` and passes it to `take_foo` that expects a `Foo` from `mls-rs-core 0.18.0`, and these 2 types are considered different, so my crate fails to build.

A similar problem can happen with traits. Does that clarify what the problem is?

mgeisler: Thankt, that explanation makes it clear! I had not considered this case with a duplication in the transitive dependencies.

mulmarta (approved):

stefunctional (approved):

PR #125, opened 2024-03-19
[uniffi] Use mls-rs-crypto-rustcrypto by default

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

It is still possible to use mls-rs-crypto-openssl if one prefers this. This is controlled with two new Cargo features: “rustcrypto” and “openssl”.

I’m not sure if this is the best way to do this kind of compile-time switching, but this seems to be fairly simple while also working in my testing.

I’m of course happy to rename things if people have better ideas for the feature names!

### DIFF
--- a/mls-rs-uniffi/src/config.rs
+++ b/mls-rs-uniffi/src/config.rs
@@ -6,7 +6,6 @@ use mls_rs::{
     identity::basic,
     storage_provider::in_memory::InMemoryGroupStateStorage,
 };
-use mls_rs_crypto_openssl::OpensslCryptoProvider;
 
 use self::group_state::{GroupStateStorage, GroupStateStorageAdapter};
 use crate::Error;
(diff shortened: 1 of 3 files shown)


mls-rs-uniffi/src/config.rs
```diff
@@ -53,10 +52,15 @@ impl mls_rs_core::group::GroupStateStorage for ClientGroupStorage {
     }
 }
 
+#[cfg(feature = "rustcrypto")]
+pub(crate) type UniFFICryptoProvider = mls_rs_crypto_rustcrypto::RustCryptoProvider;
+#[cfg(feature = "openssl")]
+pub(crate) type UniFFICryptoProvider = mls_rs_crypto_openssl::OpensslCryptoProvider;
+
```
mulmarta, line 59: Won't this crash if both features are on? IIUC Rust features should be "additive"?

PR #127, opened 2024-03-20
[uniffi] Expose `RosterUpdate` on `Commit`

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

More infrastructure to let us talk about roster updates on a group.

### Call-outs:

The deep clones are potentially worry-some here. The clone in `signing_identity` is also annoying. I don't know if it'll be a problem yet, but because we cannot return borrowed data across the FFI boundary, this is probably the best we can do right now.

### Testing:

New integration test.

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -198,6 +198,51 @@ impl From<mls_rs::group::proposal::Proposal> for Proposal {
     }
 }
 
+/// Update of a member due to a commit.
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct MemberUpdate {
+    pub prior: Arc<SigningIdentity>,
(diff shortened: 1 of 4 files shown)


mgeisler:
See https://github.com/mozilla/uniffi-rs/issues/2043 for the problem of exposing `PartialEq` and friends via UniFFI — I don't think it's possible right now, but I would love to be proven wrong :smile:

mgeisler:
> See [mozilla/uniffi-rs#2043](https://github.com/mozilla/uniffi-rs/issues/2043) for the problem of exposing `PartialEq` and friends via UniFFI — I don't think it's possible right now, but I would love to be proven wrong 😄

This is fixed, thank to the help of @mhammond in the linked issue! :tada:

stefunctional (approved):

tomleavy (approved):

PR #128, opened 2024-03-20
[uniffi] Remove `GroupState` type from storage trait

mgeisler:
### Issues:

Addresses #81 

### Description of changes:

This removes the `GroupState` type from the `GroupStateStorage` trait in mls-rs-uniffi. This is done for two reasons:

- It makes all methods on the trait consistent in the way that they take a `group_id` as the first argument.

- Fewer types means less boiler-plate generated by the UniFFI bindings generator (`Vec<u8>` is a built-in type in UniFFI).

### Call-outs:

The `EpochRecord` type stays: I don’t think we have a different way of passing in a vector of epochs IDs with their data.

I could apply this change to mls-rs too, but it doesn’t seem useful there since it’s cheap and easy to create a new type in Rust.

### Testing:

Updated tests.

### DIFF
--- a/mls-rs-uniffi/src/config.rs
+++ b/mls-rs-uniffi/src/config.rs
@@ -42,7 +42,8 @@ impl mls_rs_core::group::GroupStateStorage for ClientGroupStorage {
         updates: Vec<mls_rs_core::group::EpochRecord>,
     ) -> Result<(), Self::Error> {
         self.0.write(
-            state.into(),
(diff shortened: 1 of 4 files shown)


mgeisler:
I don't think this is super important for us, it was just something that came up when Anh was looking at the storage trait.

mulmarta (approved):

tomleavy (approved):

PR #129, opened 2024-03-22
[uniffi] Use a single welcome message in our API

mgeisler:
### Issues:

Addresses #81 

### Description of changes:

We have an expectation that all participants implement the functionalities the way we do (but not necessarily using mls-rs-uniffi as the underlying library). So if we make the assumption that `single_welcome_message = true` is always set, we can simplify the API.

### Call-outs:

I'm not super sure what the implication is of throwing away the rest of the welcome messages if someone should send us more than one?

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -315,10 +315,10 @@ impl Client {
         let basic_credential = BasicCredential::new(id);
         let signing_identity =
             identity::SigningIdentity::new(basic_credential.into_credential(), public_key.into());
-        let mls_rules = mls_rules::DefaultMlsRules::new().with_commit_options(
-            mls_rules::CommitOptions::default()
-                .with_ratchet_tree_extension(client_config.use_ratchet_tree_extension),
-        );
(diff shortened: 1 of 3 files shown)


tomleavy:
Hey @mgeisler you always receive exactly one welcome message. Generating multiple is done if you want to split to make a welcome message for each new member to save on bandwidth on the download side. The server would figure out how to fan them out to individuals separately.

tomleavy (approved):

mulmarta (approved):

stefunctional (approved):

mulmarta:
> Hey @mgeisler you always receive exactly one welcome message. Generating multiple is done if you want to split to make a welcome message for each new member to save on bandwidth on the download side. The server would figure out how to fan them out to individuals separately.

In fact it's possible (I'd have to check with the RFC) that you can anyway take the single welcome message and split it up yourself into multiple welcome messages for different joiners, each smaller than the original one. I don't think there's any crypto binding all the pieces of the single welcome message together.

mgeisler:
Thanks @tomleavy and @mulmarta for explaining!

PR #130, opened 2024-03-22
[uniffi] Reintroduce async support for UniFFI

mgeisler:
### Issues:

Addresses #81

### Description of changes:

UniFFI added support for async traits in https://github.com/mozilla/uniffi-rs/pull/1981, but this is not yet released. This commit temporarily introduces a Git dependency on UniFFI to let us test the async functionality which was removed in #86.

An unrelated UniFFI change (https://github.com/mozilla/uniffi-rs/pull/1840) made our existing test fail.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT license.

PR #131, opened 2024-03-25
Fix comment regarding macro order

stefunctional:
[Follow-up](https://github.com/awslabs/mls-rs/pull/130#discussion_r1537815777) for #130.

### DIFF
--- a/mls-rs-uniffi/src/config/group_state.rs
+++ b/mls-rs-uniffi/src/config/group_state.rs
@@ -31,9 +31,10 @@ impl From<EpochRecord> for mls_rs_core::group::EpochRecord {
     }
 }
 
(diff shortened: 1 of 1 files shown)


tomleavy (approved):

mulmarta (approved):

mgeisler:
Thanks for fixing this.

It would be nice to know _why_ this is... perhaps one of the two crates can be changed so that the order can stay the same in the two cases. It's not super important now, but I hope to look into this eventually.

stefunctional:
> It would be nice to know why this is

[Apparently](https://github.com/mozilla/uniffi-rs/pull/1981#discussion_r1482139339) `uniffi` can't deal with the lifetimes in the code generated by `async_trait`.

PR #133, opened 2024-04-02
[uniffi] Setup CI for the Kotlin code

### LINKED ISSUE #103: [uniffi] Setup CI for Kotlin
### Problem:

#101 highlighted a need for CI on the generated Kotlin code: the UniFFI generator accidentally shadowed an enum variant, and my attempt at renaming the variant did not work because a trailing `_` is also stripped out.

### Solution:

Have a GitHub CI step that tries compiling the generated Kotlin code. This would be similar to how we compile the generated Python code today.

mgeisler:
### Issues:

Resolves #103.

### Description of changes:

This sets up a simple CI pipeline for the generated Kotlin code. This would have allowed us to find and fix #101 and mozilla/uniffi-rs#2032 sooner.

### Call-outs:

Note that this updates the UniFFI dependency to the latest commit on `main` (which includes mozilla/uniffi-rs#2032 to fix the `Error` inheritance loop).

### Testing:

I still need to update the CI a bit to make the tests pass in all configurations.

### DIFF
--- a/.github/workflows/install-kotlin-deps/action.yml
+++ b/.github/workflows/install-kotlin-deps/action.yml
@@ -1,0 +1,27 @@
+# This action downloads the Kotlin dependencies using Maven. This
+# fails on Windows for some unknown reason:
+#
+#   The goal you specified requires a project to execute but there is
+#   no POM in this directory
+#
+# For now, it seems okay to just run the tests on Linux and macOS.
+name: Install Kotlin dependencies
+description: Install Kotlin dependencies for UniFFI generated code.
+runs:
+  using: composite
+  steps:
+    - name: Install JNA
+      run: |
+        mvn --no-transfer-progress dependency:copy \
+          -Dartifact=net.java.dev.jna:jna:5.14.0 \
+          -Dmdep.stripVersion=true -DoutputDirectory=$PWD
+        echo "CLASSPATH=$PWD/jna.jar:$CLASSPATH" >> "$GITHUB_ENV"
+      shell: bash
+
+    - name: Install Kotlin Test
+      run: |
+        mvn --no-transfer-progress dependency:copy \
+          -Dartifact=org.jetbrains.kotlin:kotlin-test:1.9.23 \
+          -Dmdep.stripVersion=true -DoutputDirectory=$PWD
+        echo "CLASSPATH=$PWD/kotlin-test.jar:$CLASSPATH" >> "$GITHUB_ENV"
+      shell: bash
--- a/.github/workflows/native_build.yml
+++ b/.github/workflows/native_build.yml
@@ -7,7 +7,7 @@ jobs:
   SyncBuildAndTestS:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-13, windows-latest]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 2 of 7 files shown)


mulmarta:
> Hi @mulmarta and @tomleavy, any chance that we could [configure the workflow with _Require approval for first-time contributors_](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-required-approval-for-workflows-from-public-forks)? My guess is that the setting is set to _Require approval for all outside collaborators_ right now.
> 
> Since I'm no longer a first-time contributor to this repository, it should let the workflows run immediately when I create a PR — instead of having one of you approve them first.

We tried after your first contributions. If I understand correctly, this is a per-organization and not a per-repo setting (as your link suggests) and any settings for awslabs is beyond our control.

mgeisler:
> We tried after your first contributions. If I understand correctly, this is a per-organization and not a per-repo setting (as your link suggests) and any settings for awslabs is beyond our control.

I see! Thanks for trying :smile: You've been super responsive on the PRs, so it's not a big deal for me to wait for the tests to trigger.

.github/workflows/install-kotlin-deps/action.yml
```diff
@@ -0,0 +1,31 @@
+# This action downloads the Kotlin dependencies directly from Maven
+# Central. We should to be able to use Maven for this, and the
+# following works fine on Linux and macOS:
+#
+# mvn --no-transfer-progress dependency:copy \
+#   "-Dartifact=net.java.dev.jna:jna:5.14.0" \
+#   -Dmdep.stripVersion=true -DoutputDirectory=$PWD
+#
+# However, it fails on Windows, saying
```
mgeisler, line 9: Despite my best attempts, the Windows tests now fail with a much simpler error:

```text
   Compiling mls-rs-uniffi v0.1.0 (D:\a\mls-rs\mls-rs\mls-rs-uniffi)
    Finished dev [unoptimized + debuginfo] target(s) in 26.04s
test simple_scenario_sync ... FAILED

failures:

---- simple_scenario_sync stdout ----
Creating testing out_dir: D:\a\mls-rs\mls-rs\target\tmp\mls-rs-uniffi-3416e5e0eb569dae
Error: Failed to spawn `kotlinc` to compile the bindings

Caused by:
    program not found


failures:
    simple_scenario_sync

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 43.62s
```

According to [Windows2022-Readme.md](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md), Kotlin 1.9.23 should be installed. I can also see it working:

```
$ which kotlinc
/c/tools/kotlinc/bin/kotlinc
$ kotlinc -version
info: kotlinc-jvm 1.9.23 (JRE 1.8.0_402-b06)
```

The UniFFI test somehow doesn't see it — I looked briefly at [test.rs](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/test.rs), but I don't see it overriding the `$PATH` anywhere.

I would suggest disabling the tests on Windows for now. What do people here think?

mgeisler:
Ah, the `kotlin-exception-loop` branch was missing in my clone... because mozilla/uniffi-rs#2032 was merged yesterday. Let me update this PR to use the `uniffi-rs` repository instead then!

.github/workflows/install-kotlin-deps/action.yml
tomleavy: I'm fine with disabling windows Kotlin

mgeisler: Thanks, I'll refactor the PR to do that!

mulmarta (approved):

tomleavy (approved):

PR #139, opened 2024-04-03
[uniffi] Rename `commit` to `key_update`

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

When discussing this internally, the feeling was that `commit` would mostly be used to refresh keys. The name `key_update` would thus be more descriptive.

### Call-outs:

Since the method can also be used to commit received proposals (if/when we support them), the current name might be better?

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -612,14 +612,19 @@ impl Group {
         group.export_tree().try_into()
     }
 
-    /// Perform a commit of received proposals (or an empty commit).
+    /// Perform a commit to update encryption keys.
     ///
-    /// TODO: ensure `path_required` is always set in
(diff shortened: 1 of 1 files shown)


mgeisler:
> Since the method can also be used to commit received proposals (if/when we support them), the current name might be better?

The changes in the tests show this a bit, but I'll let people here discuss :smile:

mulmarta:
I'm not so sure about this one...
1. This could have 2 purposes: PCS as key_update indicates or committing received proposals
2. Depending on your commit_options and received proposals, this may **NOT** update any keys. E.g. if you received add proposals and your commit_options don't enforce path, then this will end up being a partial commit i.e. it doesn't update any keys

beurdouche:
Without having looked into this much, a side note is that we should avoid non standard terminology as much as possible.

mgeisler:
Thanks for the feedback! Let me close this for now since it doesn't sound like a good idea.

PR #147, opened 2024-04-15
[uniffi] Introduce a `GroupId` newtype wrapper

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

This case up during our API review: instead of exposing a `Vec<u8>`, we can let the API use a `GroupId` newtype wrapper.

### Call-outs:

A downside of this is that UniFFI doesn’t support exposing traits like `Hash` on its record types (see
https://github.com/mozilla/uniffi-rs/issues/2043). A consequence of this is that the Python `GroupId` type is not hashable. I therefore ended up accessing the `.id` attribute directly in the integration test.

### DIFF
--- a/mls-rs-uniffi/src/lib.rs
+++ b/mls-rs-uniffi/src/lib.rs
@@ -337,6 +337,37 @@ pub async fn generate_signature_keypair(
     })
 }
 
+/// A group ID.
+#[derive(Clone, Debug, Eq, PartialEq, Hash, uniffi::Record)]
+pub struct GroupId {
+    id: Vec<u8>,
+}
+
+impl GroupId {
+    fn new(group_id: Vec<u8>) -> Self {
+        GroupId { id: group_id }
+    }
+}
+
+impl std::ops::Deref for GroupId {
+    type Target = [u8];
+    fn deref(&self) -> &[u8] {
+        &self.id
+    }
+}
+
+impl From<&[u8]> for GroupId {
+    fn from(group_id: &[u8]) -> Self {
+        GroupId::new(group_id.to_vec())
+    }
+}
+
+impl From<GroupId> for Vec<u8> {
+    fn from(group_id: GroupId) -> Vec<u8> {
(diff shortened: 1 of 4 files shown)


mls-rs-uniffi/src/lib.rs
```diff
@@ -408,12 +439,12 @@ impl Client {
...
     /// [`mls_rs::Client::create_group_with_id`] for details.
-    pub async fn create_group(&self, group_id: Option<Vec<u8>>) -> Result<Group, Error> {
+    pub async fn create_group(&self, group_id: Option<GroupId>) -> Result<Group, Error> {
         let extensions = mls_rs::ExtensionList::new();
         let inner = match group_id {
             Some(group_id) => {
                 self.inner
-                    .create_group_with_id(group_id, extensions)
+                    .create_group_with_id(group_id.into(), extensions)
                     .await?
```
mulmarta, line 448: We have group id bytes in a bunch of places in the main crate -- would it make sense to make it a type there as well, at least in the public API? A quick search gives `create_group_with_id(group_id, ...)`, `GroupState::id`, `GroupContext::group_id(&self)`, `CommitBuilder::reinit(group_id, ...)`

mgeisler: Yes, I think it would — I'm happy to push the changes down to that level as well. The type can be a `struct GroupId(Vec<u8>)` there — UniFFI doesn't support record-structs without field names.

mgeisler: Would you want me to push the change further down in the API stack in this PR?

tomleavy:
@mgeisler can we close this one given you will not be updating the uniffi crate anymore? I think we also want to just abandon that project overall if no one is using it.

mgeisler:
Yes, let me close it!

Removing the code again seems best now that we're pursuing an approach using Diplomat. I'm sorry we changed direction like this, but I learnt a lot from using UniFFI.

PR #149, opened 2024-04-17
[uniffi] Auto-download Kotlin dependencies

mgeisler:
### Issues:

Addresses #103.

### Description of changes:

This is a follow up to #133. It makes `cargo test` automatically download the necessary jar files used by the test code. The jars are cached in `$CARGO_TARGET_TMPDIR`, i.e., in `target/tmp`.


mgeisler:
This is based on the suggestion by @mulmarta in #133.

mulmarta (approved):
Thanks a lot! I (for one) usually run `cargo test` on workspace and this makes my life way easier.

In general, is it a common practice in Rust to list not-so-obvious tools needed to build & test? I mean kotlin and maven, but also openssl required by `mls-rs` tests.

mgeisler:
> Thanks a lot! I (for one) usually run `cargo test` on workspace and this makes my life way easier.

Yeah, mine too :smile: it was annoying to download and maintain those jars.

> In general, is it a common practice in Rust to list not-so-obvious tools needed to build & test? I mean kotlin and maven, but also openssl required by `mls-rs` tests.

I don't know of a convention — one could have a `build.rs` script which errors out with a nice message, but that's the only thing I can think of.

tomleavy (approved):

PR #150, opened 2024-04-17
Update aws-lc

mulmarta:
The version mismatch with aws-lc-sys (again) causes fails on windows https://github.com/awslabs/mls-rs/actions/runs/8723361360/job/23931412208


stefunctional (approved):

tomleavy:
@mulmarta can you bump the crate version up as well so we can just release this asap?

stefunctional (approved):

tomleavy (approved):

PR #153, opened 2024-05-02
Cryptokit clippy

mulmarta:
Add clippy and fmt CI jobs for the `mls-rs-crypto-cryptokit` package. Fix clippy warnings to make CI pass again.

##$ Call-out
CI fails because it can't create a coverage report, which should be investigated independently.

### DIFF
--- a/.github/workflows/native_build.yml
+++ b/.github/workflows/native_build.yml
@@ -104,7 +104,7 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Rust Fmt
-        run: cargo fmt -p mls-rs-crypto-cryptokit
+        run: cargo fmt --check -p mls-rs-crypto-cryptokit
       - name: Clippy
         run: cargo clippy -p mls-rs-crypto-cryptokit -- -D warnings
   CodeCoverage:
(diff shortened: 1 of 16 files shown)


.github/workflows/native_build.yml
```diff
@@ -88,6 +88,25 @@ jobs:
...
+      - uses: arduino/setup-protoc@v2
+        with:
+          version: "25.x"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      - name: Rust Fmt
+        run: cargo fmt -p mls-rs-crypto-cryptokit
```
stefunctional, line 107:
```suggestion
        run: cargo fmt --check -p mls-rs-crypto-cryptokit
```
I think `--check` needs to be passed to ask cargo-fmt to validate that the formatting is already correct instead of just formatting files that aren't right.

stefunctional:
The CI issue should be resolved by triggering a build after `safer-ffi` releases an update including getditto/safer_ffi#211.

mrdowns (approved):

PR #157, opened 2024-05-08
[uniffi] Make mls-rs-crypto-openssl optional

mgeisler:
### Issues:

Addresses #81.

### Description of changes:

For Android, we will be using a crypto provider based on BoringSSL. It is therefore important that mls-rs-uniffi becomes agnostic to the crypto provider.

This PR handles this very simply: by requiring two things:

- the user sets a Cargo feature to depend on the right crypto provider implementation,
- that the crypto provider has a suitable `new()` method that can be used to instantiate it.

fail because the `CryptoProvider` trait has an associated type, which I cannot specify at this point.

### Call-outs:

A better alternative would be to make the choice of crypto provider dynamic via a trait. Unfortunately, I don’t think this is supported with UniFFI because it doesn’t support associated types in exported traits and because it also doesn’t support generic exported types.

Making a single non-generic type which uses

```rust
struct CryptoProviderAdapter {
    crypto_provider: Box<dyn CryptoProvider>
}
```

doesn't compile because I need to specify the `CipherSuiteProvider` associated type. I don't know how I can do this in a non-generic way?

### DIFF
--- a/mls-rs-uniffi/src/config.rs
+++ b/mls-rs-uniffi/src/config.rs
@@ -6,9 +6,8 @@ use mls_rs::{
     identity::basic,
     storage_provider::in_memory::InMemoryGroupStateStorage,
 };
-use mls_rs_crypto_openssl::OpensslCryptoProvider;
 
-use self::group_state::{GroupStateStorage, GroupStateStorageAdapter};
+use crate::config::group_state::{GroupStateStorage, GroupStateStorageAdapter};
 use crate::Error;
 
 pub mod group_state;
@@ -56,10 +55,19 @@ impl mls_rs_core::group::GroupStateStorage for ClientGroupStorage {
     }
 }
 
+#[cfg(not(any(feature = "openssl")))]
+compile_error!(
+    "The crypto provider must be set by enabling exactly one of the \
+     following Cargo features: [\"openssl\"]."
+);
+
+#[cfg(feature = "openssl")]
+pub type UniFFICryptoProvider = mls_rs_crypto_openssl::OpensslCryptoProvider;
+
 pub type UniFFIConfig = client_builder::WithIdentityProvider<
     basic::BasicIdentityProvider,
     client_builder::WithCryptoProvider<
-        OpensslCryptoProvider,
+        UniFFICryptoProvider,
         WithGroupStateStorage<ClientGroupStorage, client_builder::BaseConfig>,
     >,
 >;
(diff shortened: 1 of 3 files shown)


mls-rs-uniffi/src/config.rs
```diff
@@ -56,10 +55,19 @@ impl mls_rs_core::group::GroupStateStorage for ClientGroupStorage {
     }
 }
 
+#[cfg(not(any(feature = "openssl")))]
+compile_error!(
+    "The crypto provider must be set by enabling exactly one of the \
+     following Cargo features: [\"openssl\"]."
+);
+
```
mulmarta, line 63: Would it work to make it like this?
```
#[cfg(not(any(feature = "openssl")))]
pub type UniFFICryptoProvider = mls_rs::client_builder::Missing;
```

mls-rs-uniffi/src/config.rs
```diff
@@ -56,10 +55,19 @@ impl mls_rs_core::group::GroupStateStorage for ClientGroupStorage {
...
+#[cfg(feature = "openssl")]
+pub type UniFFICryptoProvider = mls_rs_crypto_openssl::OpensslCryptoProvider;
+
```
mulmarta, line 66: If we create another feature "boringssl", we won't be able to have `UniFFICryptoProvider` defined twice, but features should be additive. Can we achieve additive features by defining `#[cfg(feature = "openssl")] UniFFIConfigOpenssl = ...` and `#[cfg(feature = "boringssl")] UniFFIConfigBoringssl = ...`?

mgeisler: > If we create another feature "boringssl", we won't be able to have `UniFFICryptoProvider` defined twice, but features should be additive.

Yes, ideally... I was going to guard against having multiple features enabled by letting the build fail with `compile_fail!` like you saw above. It's not great, but I don't know of a way to make the crypto provider dynamic at runtime.

What we need is

- a non-generic struct which holds a `mls_rs::Client` inside it
- the `mls_rs::Client` value should be able to hold different `CryptoProvider`s, implying that we should have a `CryptoProviderWrapper` type which `impl CryptoProvider` and which holds a `Box<dyn CryptoProvider>`

So I'm looking for a way to do

```rust
struct CryptoProviderWrapper {
    crypto_provider: Box<dyn CryptoProvider>,
}
```

However, this doesn't compile because I need to specify the `CipherSuiteProvider` associated type. I don't know how I can do this in a non-generic way?

mgeisler:
Let me close this again, we're now looking at using Diplomat instead of UniFFI.

PR #161, opened 2024-06-14
[crypto-rustcrypto]: remove unused custom feature

mgeisler:
### Description of changes:

We don’t actually use the “custom” feature of the getrandom crate. We don't have the feature enabled in all the places where I'm trying to use mls-rs, so it would be nice if this crate doesn't require it.

### DIFF
--- a/.github/workflows/no_std_build.yml
+++ b/.github/workflows/no_std_build.yml
@@ -26,4 +26,4 @@ jobs:
     - name: Build MLS Embedded Full RFC Compliance
       run: cargo +nightly build --package mls-rs --lib --no-default-features --features rfc_compliant --target thumbv6m-none-eabi
     - name: Build rust crypto embedded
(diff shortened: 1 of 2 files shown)


mulmarta:
Seems to be used by embedded build @mgeisler. Maybe you could add this?
```
[target.'cfg(not(feature = "std"))'.dependencies]
getrandom = { version = "0.2", default-features = false, features = ["custom"] }
```

mgeisler:
Thanks @mulmarta for the debugging! I finally came back to this and I think I understand what the problem is. Based on reading [`register_custom_getrandom!`](https://docs.rs/getrandom/latest/getrandom/macro.register_custom_getrandom.html), I now believe it was wrong for the `mls-rs-crypto-rustcrypto` _library_ to enable this feature.

It should instead be enabled by the final binary. I've updated the CI script to do this and I hope it'll work now.

mulmarta (approved):

stefunctional (approved):

PR #175, opened 2024-07-17
Remove StateUpdate in favor of CommitEffect

### LINKED ISSUE #30: Merge ProvisionalState and StateUpdate
We currently have a build flag to reduce size for embedded systems called "state_update", which eliminates the reporting of changes made as the result of a commit. There are currently ~75 references to this as a cfg flag. I propose that we instead return the existing private type `ProvisionalState` which contains all the source information required. 

1. `StateUpdate` already has all internal fields private and has functions to produce values. We can derive these from the `ProvisionalState` on demand.

2. Since everything would be calculated on demand, simply not calling these functions would produce the same binary size savings as having the feature today.

tomleavy:
### Issues:

Resolves #30 

### Description of changes:

As mentioned in #30 there are a bunch of downsides to the way that StateUpdate works currently. Upon further evaluation, the API of `StateUpdate` is not desirable to keep by simply merging it with `ProvisionalState`. It makes more sense to expose the raw `ProposalInfo` objects that were committed to (or not committed to) so that developers can use that metadata downstream. For example, if a commit has proposals from multiple authors, it is very helpful to know specifically where each proposal that was committed to originated. 

We also had a pattern of having a StateUpdate with "active" set to true or false that was somewhat overloaded and did not make sense from an idiomatic Rust perspective. The boolean is now replaced by a `CommitEffect` enum at the top level to let the caller know that a commit resulted in the local client being removed from the group, the group being reinitialized, or a new epoch. 

### Call-outs:

This is a breaking change so versions were bumped up accordingly. I tried to adapt the uniffi wrappers the best that I could, but they are already lacking in details enough that it proved to be difficult. If we are really going to support uniffi long term in this main repo we should address missing APIs.

### Testing:

All existing tests were adjusted, except for one that was removed due to it not really being necessary after the changes.


mls-rs-identity-x509/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-identity-x509"
-version = "0.11.0"
+version = "0.11.1"
```
mulmarta, line 3: As above

mls-rs-uniffi/src/lib.rs
```diff
@@ -30,7 +30,7 @@ use std::sync::Mutex;
 use tokio::sync::Mutex;
 
 use mls_rs::error::{IntoAnyError, MlsError};
-use mls_rs::group;
+use mls_rs::group::{self};
```
mulmarta, line 33:
```suggestion
use mls_rs::group;
```

mls-rs/src/client.rs
```diff
@@ -893,6 +893,8 @@ mod tests {
         // interim_transcript_hash to be computed from the confirmed_transcript_hash and
         // confirmation_tag, which is not the case for the initial interim_transcript_hash.
 
+        use crate::group::{message_processor::CommitEffect, CommitMessageDescription};
+
```
mulmarta, line 897: [Nit] why here and not on top?

mls-rs/src/group/message_processor.rs
```diff
@@ -776,34 +630,38 @@ pub(crate) trait MessageProcessor: Send + Sync {
...
         }
 
+        let commit_effect =
+            if let Some(reinit) = provisional_state.applied_proposals.reinitializations.pop() {
+                self.group_state_mut().pending_reinit = Some(reinit.proposal.clone());
+                CommitEffect::ReInit(reinit)
+            } else {
+                CommitEffect::NewEpoch(Box::new(NewEpoch {
+                    epoch: provisional_state.group_context.epoch,
+                    prior_state: self.group_state().clone(),
```
mulmarta, line 655: Intuitively, we should be able to move out from the current state instead of clone, since we're going to replace it by the provisional one? We're cloning a large thing here. Can be a follow-up, maybe it's a bigger change.

mls-rs/src/group/state.rs
```diff
@@ -2,25 +2,52 @@
...
 #[derive(Clone, Debug, PartialEq)]
 #[non_exhaustive]
 pub struct GroupState {
     #[cfg(feature = "by_ref_proposal")]
     pub(crate) proposals: crate::group::ProposalCache,
-    pub(crate) context: GroupContext,
+    pub context: GroupContext,
     pub(crate) public_tree: TreeKemPublic,
     pub(crate) interim_transcript_hash: InterimTranscriptHash,
     pub(crate) pending_reinit: Option<ReInitProposal>,
```
mulmarta, line 28: Should this be accessible to applications too?

mulmarta (dismissed):
Awesome, that simplifies so much! Left some comments, mostly nits.

mls-rs-crypto-hpke/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-hpke"
-version = "0.9.0"
+version = "0.9.1"
```
stefunctional, line 3: Should this be `0.10.0` given the breaking update of `mls-rs-core`?

mls-rs-provider-sqlite/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-provider-sqlite"
-version = "0.12.0"
+version = "0.12.1"
```
stefunctional, line 3: `0.13.0`?

mls-rs/src/group/message_processor.rs
```diff
@@ -102,73 +81,48 @@ pub(crate) fn path_update_required(proposals: &ProposalBundle) -> bool {
...
-    pub fn pending_reinit_ciphersuite(&self) -> Option<CipherSuite> {
-        self.pending_reinit
-    }
+#[cfg_attr(
+    all(feature = "ffi", not(test)),
+    safer_ffi_gen::ffi_type(clone, opaque)
+)]
+#[derive(Clone, Debug, PartialEq)]
+pub enum CommitEffect {
+    NewEpoch(Box<NewEpoch>),
```
stefunctional, line 123: Out of curiosity, did clippy suggest to box it? Just asking because it makes pattern matching more difficult.

stefunctional (approved):

mls-rs/src/group/state.rs
tomleavy: I eventually want to make mostly all of this public but it felt like that should be a separate PR later. Since those changes are only additive it could be done in a patch release.

mls-rs/src/group/message_processor.rs
tomleavy: Yeah, you have to Box this because the difference in size is very large. Something like 600 bytes

mls-rs/src/client.rs
tomleavy: Only used if the psk feature is on

mls-rs/src/group/message_processor.rs
stefunctional: Got it. Thank you.

stefunctional (approved):

mulmarta (approved):

mls-rs/src/group/message_processor.rs
tomleavy: It would definitely be a larger change since we don't make the swap until later when every succeeds. If you do a swap here and then there is an error between that and the end of processing you would wind up with a corrupted state

mulmarta: Yes, I was thinking that we could set this value as the very last thing, but that's hard with the way `update_key_schedule` makes the final changes, and it's a trait function... Would be nice to have at some point, maybe?

PR #176, opened 2024-07-19
Add RemoveProposal to the Removed case of CommitEffect

tomleavy:
### Description of Changes:

The removed case of CommitEffect didn't provide any details about who removed you, so we should provide the proposal that triggered your removal

### DIFF
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -73,7 +73,6 @@ pub(crate) use self::commit::test_utils::CommitModifiers;
 #[cfg(all(test, feature = "private_message"))]
 pub use self::framing::PrivateMessage;
 
-#[cfg(feature = "psk")]
 use self::proposal_filter::ProposalInfo;
 
 #[cfg(any(feature = "secret_tree_access", feature = "private_message"))]
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -1827,13 +1827,19 @@ where
         &mut self.state
     }
 
-    fn can_continue_processing(&self, provisional_state: &ProvisionalState) -> bool {
-        !(provisional_state
-            .applied_proposals
-            .removals
-            .iter()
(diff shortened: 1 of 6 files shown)


mulmarta (approved):

mls-rs/src/group/mod.rs
```diff
@@ -1827,13 +1826,19 @@ where
...
-            .applied_proposals
-            .removals
-            .iter()
-            .any(|p| p.proposal.to_remove == self.private_tree.self_index)
-            && self.pending_commit.is_none())
+    fn removal_proposal(
+        &self,
+        provisional_state: &ProvisionalState,
+    ) -> Option<ProposalInfo<RemoveProposal>> {
+        match self.pending_commit {
```
stefunctional, line 1833: It's not clear to me why this test is done.

stefunctional (dismissed):

mls-rs/src/group/mod.rs
tomleavy: It wasn't clear to me either, but it is there to ensure that external commits work. Maybe @mulmarta has a better idea of why this works, because I forget

mulmarta: I think for external commit, `private_tree.self_index = 0` at this point because it has to be something, but you don't know what it should be before you apply the proposals. So if the external committer removes its old copy, and the old copy has index `0`, this would make the external committer fail.

I will have to double check though.

mulmarta (approved):

mls-rs/src/group/mod.rs
stefunctional: Thank you for the answers. How are these indexes related to checking if there's a pending commit?

stefunctional (approved):

mls-rs/src/group/mod.rs
tomleavy: Ok @mulmarta I think if there is some issue there or better way to detect this situation we can patch it after as what is there now is equal to what was there before. Let us know if you find anything out.

PR #183, opened 2024-08-16
Let `TestGroup` deref to `Group`

mgeisler:
### Description of changes:

This lets us treat a `TestGroup` as a `Group` in most places and thus turn a lot of `alice_group.group` lines into just `alice_group`.

### Call-outs:

This makes the code less explicit, but on balance I like the new way since it reduces stutter.

### Testing:

Unit tests pass before/after.


mulmarta (approved):
Awesome, thanks!

stefunctional (approved):

PR #184, opened 2024-08-16
Add support for last resort key packages

mgeisler:
### Description of changes:

This introduces a `LastResortKeyPackageExt` extension which corresponds to the last resort key package extension described here:

  https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions-04#name-last-resort-keypackages

The new code is hidden behind a `last_resort_key_package_ext` Cargo feature.

There is not yet a super convenient way to set the extension on a single key package: the key package extensions are fixed when the client is created, so one has to create a temporary throw-away client for this. This can be addressed in a future PR, if needed.

### Call-outs:

I created the new `LastResortKeyPackageExt` type for consistency with the other extensions — but it's not strictly needed since the extension carries no data.

### Testing:

New tests were added.

### DIFF
--- a/mls-rs/Cargo.toml
+++ b/mls-rs/Cargo.toml
@@ -32,6 +32,7 @@ by_ref_proposal = []
 psk = []
 x509 = ["mls-rs-core/x509", "dep:mls-rs-identity-x509"]
 rfc_compliant = ["private_message", "custom_proposal", "out_of_order", "psk", "x509", "prior_epoch", "by_ref_proposal", "mls-rs-core/rfc_compliant"]
+last_resort_key_package_ext = ["mls-rs-core/last_resort_key_package_ext"]
 
 std = ["mls-rs-core/std", "mls-rs-codec/std", "mls-rs-identity-x509?/std", "hex/std", "futures/std", "itertools/use_std", "safer-ffi-gen?/std", "zeroize/std", "dep:debug_tree", "dep:thiserror", "serde?/std"]
 
--- a/mls-rs/src/extension.rs
+++ b/mls-rs/src/extension.rs
@@ -5,9 +5,15 @@
 pub use mls_rs_core::extension::{ExtensionType, MlsCodecExtension, MlsExtension};
 
 pub(crate) use built_in::*;
+#[cfg(feature = "last_resort_key_package_ext")]
+pub(crate) use recommended::*;
 
 /// Default extension types required by the MLS RFC.
 pub mod built_in;
+
+/// Extension types which are not mandatory, but still recommended.
+#[cfg(feature = "last_resort_key_package_ext")]
+pub mod recommended;
 
 #[cfg(test)]
 pub(crate) mod test_utils {
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -7,6 +7,8 @@ use alloc::vec::Vec;
 use core::fmt::{self, Debug};
 use mls_rs_codec::{MlsDecode, MlsEncode, MlsSize};
 use mls_rs_core::error::IntoAnyError;
+#[cfg(feature = "last_resort_key_package_ext")]
+use mls_rs_core::extension::MlsExtension;
 use mls_rs_core::secret::Secret;
 use mls_rs_core::time::MlsTime;
 
@@ -14,6 +16,8 @@ use crate::cipher_suite::CipherSuite;
 use crate::client::MlsError;
 use crate::client_config::ClientConfig;
 use crate::crypto::{HpkeCiphertext, SignatureSecretKey};
+#[cfg(feature = "last_resort_key_package_ext")]
+use crate::extension::LastResortKeyPackageExt;
 use crate::extension::RatchetTreeExt;
 use crate::identity::SigningIdentity;
 use crate::key_package::{KeyPackage, KeyPackageRef};
@@ -431,9 +435,9 @@ where
         let (encrypted_group_secrets, key_package_generation) =
             find_key_package_generation(&config.key_package_repo(), &welcome.secrets).await?;
 
-        let key_package_version = key_package_generation.key_package.version;
+        let key_package = &key_package_generation.key_package;
 
-        if key_package_version != protocol_version {
+        if key_package.version != protocol_version {
             return Err(MlsError::ProtocolVersionMismatch);
         }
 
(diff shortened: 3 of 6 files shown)


mls-rs/Cargo.toml
```diff
@@ -16,7 +16,7 @@ features = ["external_client", "sqlite"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["std", "rayon", "rfc_compliant", "tree_index", "fast_serialize"]
+default = ["std", "rayon", "rfc_compliant", "tree_index", "fast_serialize", "last_resort_key_package_ext"]
```
mgeisler, line 19: By the way, there is no particular need for this to be a default feature — I just figured that others would like to use it too. But let me take it out for simplicity.

mls-rs/src/extension.rs
```diff
@@ -9,6 +9,10 @@ pub(crate) use built_in::*;
 /// Default extension types required by the MLS RFC.
 pub mod built_in;
 
+/// Draft extension types.
+#[cfg(feature = "last_resort_key_package_ext")]
+pub mod draft;
```
mulmarta, line 14: Why `draft`?

mls-rs/src/group/mod.rs
```diff
@@ -516,10 +520,17 @@ where
...
 
-        let used_key_package_ref = key_package_generation.reference;
+        #[cfg(not(feature = "last_resort_key_package_ext"))]
+        let is_last_resort = false;
+        #[cfg(feature = "last_resort_key_package_ext")]
+        let is_last_resort = key_package
+            .extensions
+            .has_extension(LastResortKeyPackageExt::extension_type());
+        // Delete the key just used if this is not a last-resort key package.
+        let used_key_package_ref = (!is_last_resort).then_some(key_package_generation.reference);
```
mulmarta, line 533: I may be opening a can of worms here, but would it be useful to add a function on SQLite storage to delete all last resort packages? This way we could rotate them before they expire?

mls-rs/src/extension.rs
mgeisler: I am happy to call the module anything you like — I thought the extensions in [the RFC](https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions-04) are still draft extensions? Or rather that the RFC is a draft?

I called the module `proposals` at some point. I don't know if that's better?

mls-rs/src/group/mod.rs
mgeisler: Yeah, that could make sense.

I guess the behavior here is left up to the implementation by the RFC. I'm happy to extend the SQLite layer with such a method, perhaps in a later PR if people feel it's needed?

mls-rs/src/extension.rs
mulmarta: Ah I see. The RFC is a draft, yes. I thought this meant that the implementation is a draft, and got confused. Maybe something like "non-default" since the comment above `built_in` mentions "default"?

Why `proposals` is this is about key packages?

mls-rs/src/group/mod.rs
mulmarta: That would be great, but remember that we need backwards compatibility on the DB :/

mls-rs/src/extension.rs
mgeisler: > Ah I see. The RFC is a draft, yes. I thought this meant that the implementation is a draft, and got confused. Maybe something like "non-default" since the comment above `built_in` mentions "default"?

I'm not sure the distinction is unnecessary. More precisely, the extensions in the `built_in` module are the ones that have _Recommended: yes_ in [17.3. ](https://www.rfc-editor.org/rfc/rfc9420.html#section-17.3)[MLS Extension Types](https://www.rfc-editor.org/rfc/rfc9420.html#name-mls-extension-types).

The extensions in [The Messaging Layer Security (MLS) Extensions](https://datatracker.ietf.org/doc/html/draft-ietf-mls-extensions-04) are actually also recommended, but they don't have a corresponding RFC yet.

Since you already have a `pub(crate) use built_in::*`, I would actually propose to just get rid of the `built_in` module. Or perhaps it could be renamed to `recommended` if you like? Then we could add the last resort extension there.
 
> Why `proposals` is this is about key packages?

Here I meant "proposal" in the sense of "proposed extensions", but I see that it's confusing when the protocol itself uses this term :smile:

mls-rs/src/group/mod.rs
mgeisler: Given that we have [`SqLiteKeyPackageStorage::delete`](https://docs.rs/mls-rs/latest/mls_rs/storage_provider/sqlite/storage/struct.SqLiteKeyPackageStorage.html#method.delete) already, perhaps it would be more appropriate to

- add a way to iterate over the key packages
- add a method to determine if a key package is a last resort package

mgeisler: Let me resolve this for now — this PR is about the core functionality. Deleting non-expired key packages seems like a potential next step to me.

mgeisler:
I think I've traced down the small build failures now :smile: Can I get a retry of the CI pipelines?

mls-rs/src/extension.rs
mulmarta: There is a difference between "recommended" and "built-in" which I think was renamed to "default" in the RFC. See [here](https://www.rfc-editor.org/rfc/rfc9420.html#section-7.2-4). Default extensions MUST be supported by all MLS clients. Recommended extensions may or may not be supported but they're OK to support.

The extensions in the `built_in` module are "default" ones. If there is a "recommended" extensions in the Extensions RFC, then it would not be "default" and so not in `built_in`

mgeisler: Okay, would you like me to rename the module to `recommended`? Any name is fine with me :smile:

mulmarta: Yeah, I just found "draft" misleading, sounds like "not fully implemented". Recommended is fine, I don't mind as long as it's not very confusing.

mgeisler: Super, I renamed it to `recommended` now. I also added a `pub use` import like for `built_in` — I feel this is helpful since it removes the need for others to think about this detail.

mulmarta: Awesome

mulmarta (approved):

stefunctional (approved):

mgeisler:
Thanks a lot @mulmarta and @stefunctional for the review! Would you be able to make a new release with this in the next days? That will make it easier for me to start using the new functionality (we typically don't vendor the source code based on Git commits).

mulmarta:
> Thanks a lot @mulmarta and @stefunctional for the review! Would you be able to make a new release with this in the next days? That will make it easier for me to start using the new functionality (we typically don't vendor the source code based on Git commits).

Would you be willing to bump the versions? I think minor versions are sufficient since you only added a function. Then I can publish to crates.io

mgeisler:
Sure, I'm happy to send a PR! I agree with you that this is a simple backwards compatible change.

PR #186, opened 2024-08-19
Format with `imports_granularity = "Module"`

mgeisler:
### Description of changes:

This is the result of running `cargo +nightly fmt` with the unstable `rustfmt` option `imports_granularity = "Module"`.

Personally, I find this style much easier to use:

- The lack of deep nesting makes it easier for me to understand the imports at a glance.
- Because most imports are on a single line, I can add/remove them easier. Copy-pasting between files is easier.

### Call-outs:

This option is unstable, meaning that it has no effect on a stable `rustfmt`. Luckily, the default value for `imports_granularity` is `Preserve`, which means that the formatting done here will be preserved by future `cargo fmt` calls:

  https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity

This applies to both on the stable and nightly channels.

This is a personal preference, so this PR is just there to demonstrate the delta in case it hadn't been considered. Somewhat unintuitively, this style is more compact: the diffstat says `971 insertions(+), 1187 deletions(-)`, meaning that the imports take up 18% less vertical space now.

### DIFF
--- a/mls-rs-codec-derive/src/lib.rs
+++ b/mls-rs-codec-derive/src/lib.rs
@@ -4,10 +4,8 @@
 
 use std::str::FromStr;
 
-use darling::{
-    ast::{self, Fields},
-    FromDeriveInput, FromField, FromVariant,
-};
+use darling::ast::{self, Fields};
+use darling::{FromDeriveInput, FromField, FromVariant};
 use proc_macro2::{Literal, TokenStream};
 use quote::quote;
 use syn::{
--- a/mls-rs-codec/src/byte_vec.rs
+++ b/mls-rs-codec/src/byte_vec.rs
@@ -2,7 +2,8 @@
 // Copyright by contributors to this project.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-use crate::{iter::mls_decode_split_on_collection, Error, MlsEncode, MlsSize, VarInt};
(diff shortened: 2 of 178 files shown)


mulmarta (dismissed):
Cool, thanks!

For sure this makes the style consistent, which is already a big improvement.

I personally don't mind which style is used

mgeisler:
> Cool, thanks!
> 
> For sure this makes the style consistent, which is already a big improvement.

Cool, I'm glad you like it! I'll fix the fuzz target imports too now.

mgeisler:
I think we're getting closer — the fuzz targets are also formatted now. I thought they would get formatted by a `cargo fmt --all`, but I was wrong :sweat_smile:

mulmarta (approved):

stefunctional (approved):

stefunctional:
This formatting style deviates from rust-analyzer's default, but that's probably a moot point as rust-analyzer seems to recognize the style and adapts automatic imports accordingly.

This style is inconsistent for conditional imports. This PR contains several blocks looking like this:
```rust
use crate::{
    group::{
        framing::{Content, MlsMessagePayload},
        message_processor::CachedProposal,
        message_signature::AuthenticatedContent,
        proposal::Proposal,
        proposal_ref::ProposalRef,
        Sender,
    },
    WireFormat,
};
```

PR #188, opened 2024-08-27
Add GroupInfo and Welcome message decryption APIs

mulmarta:
Add APIs to decrypt group info from welcome message (without tree) and validate group info (without tree). They may be useful.

*Code in group/mod.rs was copied to a new function but not modified*

### Testing:
Added unit test

### DIFF
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -20,7 +20,7 @@ use crate::crypto::{HpkeCiphertext, SignatureSecretKey};
 use crate::extension::LastResortKeyPackageExt;
 use crate::extension::RatchetTreeExt;
 use crate::identity::SigningIdentity;
-use crate::key_package::{KeyPackage, KeyPackageRef};
+use crate::key_package::{KeyPackage, KeyPackageGeneration, KeyPackageRef};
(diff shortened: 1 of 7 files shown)


stefunctional (approved):

mls-rs/src/group/mod.rs
```diff
@@ -1617,6 +1548,145 @@ where
...
+
+    #[cfg(not(feature = "psk"))]
+    #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    async fn psk_secret<CS: CipherSuiteProvider>(
+        _config: &C,
+        cipher_suite_provider: &CS,
+        _psks: &[PreSharedKeyID],
+    ) -> Result<PskSecret, MlsError> {
+        Ok(PskSecret::new(cipher_suite_provider))
+    }
```
tomleavy, line 1597: I feel like having all this logic in group is a bad thing and we should plan on how to move it someplace more modular?

tomleavy (approved):

mls-rs/src/group/mod.rs
mulmarta: Yeah agreed

PR #189, opened 2024-09-01
feat(nodejs): add nodejs webcrypto provider through feature gates

nicdard:
### Issues:

Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes:

As NodeJs is implementing the WebCrypto API specification, adding the possibility to conditionally compile the webcrypto provider for NodeJs environment. 
This can be useful to enhance interoperability.
In my case specifically, I need to have a node compatible provider to simplify testing and benchmarking my application using this library compiled to WASM.

### Call-outs:

I am still a beginner in Rust and WASM 

### Testing:

This change is currently used in my master thesis project (@mulmarta we contacted you a while ago, David Balbás wrote you an email to ask around exporting the epoch secret).


mls-rs-crypto-webcrypto/js/node-crypto.js
```diff
@@ -0,0 +1,5 @@
+const crypto = require("node:crypto");
```
tomleavy, line 1: Does node:crypto have the exact same api as subtle crypto?

mls-rs-core/Cargo.toml
```diff
@@ -23,6 +25,7 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 
 [dependencies]
+cfg-if = "1.0.0"
```
tomleavy, line 28: this doesn't seem to be used?

mls-rs-core/Cargo.toml
```diff
@@ -11,7 +11,9 @@ exclude = ["test_data"]
 
 
 [features]
-default = ["std", "rfc_compliant", "fast_serialize"]
+web = []
+node = []
+default = ["std", "rfc_compliant", "fast_serialize", "web"]
```
tomleavy, line 16: I don't think this is needed because it is covered in mls-rs-crypto-webcrypto

mls-rs-core/src/time.rs
```diff
@@ -57,6 +57,15 @@ extern "C" {
...
 
+#[cfg(all(target_arch = "wasm32", feature = "node"))]
+#[wasm_bindgen(inline_js = r#"
+module.exports.date_now = function() {
+    return Date.now();
+};"#)]
+extern "C" {
+    fn date_now() -> f64;
+}
+
```
mulmarta, line 68: Does this mean that I can't compile this for `wasm32` with `--no-default-features` (since date_now is undefined) or `--all-features` (since date_now is defined twice)? I think in rust features should be additive. Maybe something like `not(feature = "web")` here? Not sure

mulmarta (commented):
Thanks @nicdard and hope to see the thesis / paper at some point!

mls-rs-crypto-webcrypto/js/node-crypto.js
nicdard: It seems to me, but I was re-checking the documentation and saw that the correct way is to use:
```suggestion
const crypto = require("node:crypto").webcrypto;
```
I'll test this and change accordingly

mls-rs-core/Cargo.toml
nicdard: sorry probably left there by mistake

nicdard:
> Thanks @nicdard and hope to see the thesis / paper at some point!

Thanks! sure I will share later on when I have the thesis ready if you are interested :)

mls-rs-core/src/time.rs
nicdard: oh ok I see, I will try to make them additive, thanks!

mls-rs-core/Cargo.toml
nicdard: could you elaborate?

nicdard: are you referring to this: https://doc.rust-lang.org/cargo/reference/features.html#feature-unification ?

nicdard: cool thanks for pointing out, this helped me with finding a good solution :)

mls-rs-core/src/time.rs
nicdard: I think I have implemented a good solution now

nicdard:
@mulmarta @tomleavy I found another compatibility issue with NodeJS: While Chrome generates the public key from a EC private key when importing it fro pkcs8 format, NodeJs doesn't, therefore an empty commit or an update proposal would fail. I am importing the key in jwk to force the generation of the public key part.

The same issue is also present in safari and firefox, with the difference that those are throwing an exception already when you try to import a pkcs8 key as a CryptoKey if it contains only the secret key bytes.

PR #193, opened 2024-09-23
Add support for different SQLite journal modes

JHabz:
### Issues:

N/A

### Description of changes:

Describe the current behavior and how your code changes that behavior. If there are no issues that this PR resolves, please create one first.

Allows for changing the journal mode for SQLite. This potentially also mitigates an issue where the database could get corrupted if a connection terminates while being updated from the default journal mode.

### Call-outs:

Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

N/A

### Testing:

How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

Added a unit test covering the functionality. Existing unit tests which tangentially use the function are also passing.

### DIFF
--- a/mls-rs-provider-sqlite/Cargo.toml
+++ b/mls-rs-provider-sqlite/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-provider-sqlite"
-version = "0.13.0"
+version = "0.14.0"
 edition = "2021"
 description = "SQLite based state storage for mls-rs"
 homepage = "https://github.com/awslabs/mls-rs"
--- a/mls-rs-provider-sqlite/Cargo.toml
+++ b/mls-rs-provider-sqlite/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-provider-sqlite"
-version = "0.14.0"
+version = "0.13.1"
 edition = "2021"
 description = "SQLite based state storage for mls-rs"
 homepage = "https://github.com/awslabs/mls-rs"
--- a/mls-rs-provider-sqlite/src/lib.rs
+++ b/mls-rs-provider-sqlite/src/lib.rs
@@ -193,7 +193,7 @@ fn create_tables_v1(connection: &Connection) -> Result<(), SqLiteDataStorageErro
 
 #[cfg(test)]
 mod tests {
-    use std::{fs, path::Path};
+    use std::env::temp_dir;
 
     use crate::{
         connection_strategy::{FileConnectionStrategy, MemoryStrategy},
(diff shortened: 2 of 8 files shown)


mls-rs-provider-sqlite/src/lib.rs
```diff
@@ -54,13 +54,39 @@ impl mls_rs_core::error::IntoAnyError for SqLiteDataStorageError {
...
+    DELETE,
+    TRUNCATE,
+    PERSIST,
+    MEMORY,
+    WAL,
+    OFF
+}
+
+// Note: for in-memory dbs (such as what the tests use), the only available options are MEMORY or OFF
+// Invalid modes do not error, only no-op
```
stefunctional, line 68: Would it make sense to make this a doc comment (`///`) and attach it to the definition of `JournalMode` so that it shows up in the documentation?

mls-rs-provider-sqlite/src/lib.rs
```diff
@@ -69,9 +95,11 @@ where
 {
     pub fn new(
         connection_strategy: CS,
+        journal_mode: Option<JournalMode>,
```
stefunctional, line 98: Have you considered not adding this parameter and instead having a separate function like the following?

```rust
/// A `journal_mode` of `None` means the SQLite default is used.
pub fn with_journal_mode(self, journal_mode: Option<JournalMode>) -> Self {
    Self {
        journal_mode,
        ..self
    }
}
```

This has the advantage of not being a breaking change. The current approach is a breaking change for `mls-rs-provider-sqlite` and `mls-rs`.

mls-rs-provider-sqlite/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-provider-sqlite"
-version = "0.13.0"
+version = "0.14.0"
```
stefunctional, line 3: This can be set to `0.13.1` now that the changes are non-breaking.

mls-rs-provider-sqlite/src/lib.rs
```diff
@@ -170,4 +216,29 @@ mod tests {
 
         assert_eq!(current_schema, 1);
     }
+
+    #[test]
+    pub fn journal_mode_test() {
+        let db_path = Path::new("./test_db.sqlite");
```
stefunctional, line 222: May you use [`tempdir`](https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html)? That is the usual way to create files during tests and it takes care of cleaning automatically.

mls-rs-provider-sqlite/src/lib.rs
```diff
@@ -170,4 +216,24 @@ mod tests {
 
         assert_eq!(current_schema, 1);
     }
+
+    #[test]
+    pub fn journal_mode_test() {
+        let db_path = temp_dir().join("test_db.sqlite");
```
stefunctional, line 222: As linked previously, this should use `tempdir` from the `tempfile` crate, which we already depend on. `std::env::temp_dir` has several issues:
- No clean-up is performed, or at least it is platform-dependent.
- Other programs may use it and files may conflict.

`tempdir` avoids all this trouble by creating a randomly named directory and taking care of deleting it.

mls-rs/Cargo.toml
```diff
@@ -63,7 +63,7 @@ spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin
 maybe-async = { version = "0.2.10" }
 
 # Optional dependencies
-mls-rs-provider-sqlite = { path = "../mls-rs-provider-sqlite", version = "0.13.0", default-features = false, optional = true }
+mls-rs-provider-sqlite = { path = "../mls-rs-provider-sqlite", version = "0.13.1", default-features = false, optional = true }
```
stefunctional, line 66: If this is changed (which is not necessary until `mls-rs` calls the new function), `mls-rs` version should also be incremented so that it can be published.

stefunctional (approved):
Looks good. Thank you!

tomleavy (approved):

PR #196, opened 2024-09-30
Rework leaf node extensions to work via parameters rather than as a c…

### LINKED ISSUE #195: Improve Leaf Node Extension Management
### Description of feature:

This feature makes breaking changes to the management of leaf node extensions (LNE) in mls-rs by moving LNEs outside of the Client config.

### Use case:

The current implementation of leaf node extensions (LNE) requires the Client to be initialised with the extensions, typicaly by configuring them on the Client Builder. This comes with the following drawbacks:

- Making changes to a member's LNEs requires either reinitializing the entire client or making the LNE itself a stateful reference type rather than a plain data structure. Neither approach is particularly developer friendly
- Commit operations automatically update the LNEs of the member making the commit. Whilst this is likely desirable for some extensions, it puts an undue burden on consumers who don't want to make updates to the LNEs implicitly. Such consumers must ensure they preserve the data previously configured as the LNE in some form of external storage so that it can be passed to the Client upon initialisation. Failure to do so would result in LNEs being wrongly overwritten when making commits.

- Fully remove ClientConfig.leaf_node_extensions (Breaking change.)
   - I believe breaking the existing API is necessary to ensure we avoid some of the undesirable behaviours in the discussion section
- Allow leaf_node_extensions to be passed as parameters when executing the following operations:
   - Creating a group (ie, setting the LNEs for the member that creates the group)
   - Generating a KeyPackage
   - Making Commits (via a property in CommitBuilder)
      - If the property is not passed, then the commit code should be modified to preserve whatever value is currently set in LeadNode.extensions rather than overwrite it
   - Making update proposals
       - If the property is not passed, then the commit code should be modified to preserve whatever value is currently set in LeadNode.extensions rather than overwrite it

CaioSym:
### Issues:

Resolves #195 

### Description of changes:

- Fully remove ClientConfig.leaf_node_extensions AND ClientConfig.key_package_extensions (Breaking change.)
   - I believe breaking the existing API is necessary to ensure we avoid some of the undesirable behaviours in the discussion section
- Allow leaf_node_extensions to be passed as parameters when executing the following operations:
   - Creating a group (ie, setting the LNEs for the member that creates the group)
   - Generating a KeyPackage / KeyPackage message
   - Making Commits (via a property in CommitBuilder)
      - If the property is not passed, then the commit code should be modified to preserve whatever value is currently set in LeadNode.extensions rather than overwrite it
   - Making update proposals
       - If the property is not passed, then the commit code should be modified to preserve whatever value is currently set in LeadNode.extensions rather than overwrite it 
- Allow key_package_extensions to be passed as parameters when executing the following operations:
   - Generating a KeyPackage / KeyPackage message

 
### Call-outs:

This is a breaking change as it reworks several methods in the public API. We could mitigate some of this by introducing more methods that roll out default LNEs.

### Testing:

Current unit tests already cover all of the affected areas. All tests are passing. Some tests were modified so as to leverage the new apis introduced in this PR.


mls-rs/src/client_builder.rs
```diff
@@ -297,56 +297,6 @@ impl<C: IntoConfig> ClientBuilder<C> {
...
-        extensions: ExtensionList,
-    ) -> ClientBuilder<IntoConfigOutput<C>> {
-        let mut c = self.0.into_config();
-        c.0.settings.leaf_node_extensions.append(extensions);
-        ClientBuilder(c)
-    }
-
     /// Set the lifetime duration in seconds of key packages generated by the client.
     pub fn key_package_lifetime(self, duration_in_s: u64) -> ClientBuilder<IntoConfigOutput<C>> {
         let mut c = self.0.into_config();
```
mulmarta, line 302: I wonder if this should be flexible too? E.g. longer lifetime for last resort key package?

mls-rs/src/group/commit.rs
```diff
@@ -452,11 +463,13 @@ where
             group_info_extensions: Default::default(),
             new_signer: Default::default(),
             new_signing_identity: Default::default(),
+            new_leaf_node_extensions: Default::default(),
         }
     }
 
     /// Returns commit and optional [`MlsMessage`] containing a welcome message
     /// for newly added members.
+    #[allow(clippy::too_many_arguments)]
```
mulmarta, line 472: Can we do something about it?

mls-rs/src/group/mod.rs
```diff
@@ -288,6 +288,7 @@ impl<C> Group<C>
 where
     C: ClientConfig + Clone,
 {
+    #[allow(clippy::too_many_arguments)]
```
mulmarta, line 291: As above?

mls-rs/src/group/resumption.rs
```diff
@@ -186,13 +190,18 @@ impl<C: ClientConfig + Clone> ReinitClient<C> {
             extensions: self.reinit.new_group_context_extensions(),
         };
 
+        // TODO, should these come from the reinit proposal or as an arg instead?
+        let group = self.client.load_group(self.reinit.group_id())?;
+        let current_leaf_node_extensions = &group.current_user_leaf_node()?.extensions;
+
```
mulmarta, line 196: There are no leaf node extensions in the reinit proposal -- only group context extensions... I'd say we don't support updating extensions in reinit and that's fine. Reinit is too complicated as it is.

mulmarta (dismissed):
Thanks @CaioSym ! Left a couple of nit comments.

mls-rs/src/group/resumption.rs
CaioSym: Sounds good to me. I'll remove the TODO comment then.

mls-rs/src/client_builder.rs
CaioSym: Makes sense. I'll extract it.

mls-rs/src/group/commit.rs
mulmarta: Can be a follow up.

mls-rs/src/group/mod.rs
mulmarta: Sure

mls-rs/src/client_builder.rs
CaioSym: So, after trying to extract this I ran into quite a few unforeseen issues. Despite this parameter being named `key_package_lifetime`, It seems to be used whenever we create a node in the tree, including places like group creation where a KP is not normally involved in the formal sense. This made extraction quite cumbersome.

I'm open to discussing extracting this but would push for that to be a separate issue given the above. Thoughts?

mls-rs/src/group/resumption.rs
CaioSym: Had to update this to be a parameter, Turns out load group won't work if the group has been reinited....

mls-rs/src/client_builder.rs
mulmarta: If it's used to create leaf nodes for updates, that's very strange. If you could create an issue to investigate, that would be great. Definitely not part of this PR.

mls-rs/src/group/commit.rs
```diff
@@ -560,7 +584,7 @@ where
                 &mut provisional_group_context,
                 &provisional_state.indexes_of_added_kpkgs,
                 new_signer_ref,
-                self.config.leaf_properties(),
+                Some(self.config.leaf_properties(new_leaf_node_extensions)),
```
mulmarta, line 587: This way, do we ever pass None here (other than in tests)?

mls-rs/src/group/mod.rs
```diff
@@ -851,16 +853,19 @@ where
...
 
+        let new_leaf_node_extensions =
+            leaf_node_extensions.unwrap_or(new_leaf_node.ungreased_extensions());
         let secret_key = new_leaf_node
             .update(
                 &self.cipher_suite_provider,
                 self.group_id(),
                 self.current_member_index(),
-                self.config.leaf_properties(),
+                Some(self.config.leaf_properties(new_leaf_node_extensions)),
```
mulmarta, line 868: As above - do we ever pass None?

CaioSym: At the moment no.

My original plan was to allow passing `None` so that we don't necessarily require new LNEs to be passed but given that the config is a combination of capabilities + LNEs and capabilities are currently driven by the Client's config object I opted to err on the side of preserving the current behavior of ensuring the new capabilities of the client are always set.

I don't see any harm in allowing this internal API to accept an Option but happy to revert it if we prefer.

mls-rs/src/client_builder.rs
CaioSym: Filed https://github.com/awslabs/mls-rs/issues/197

mulmarta: Thank you!

mulmarta (approved):

tomleavy (approved):

PR #198, opened 2024-10-17
Support AWS-LC in FIPS mode

mulmarta:
Changes:
* FIPS does not have `X509_get0_pubkey` so it's replaced by `X509_get_pubkey` which differs in requiring `free` of the key
* It seems that aws-lc-rs updated `BIO_number_written` and now it returns `u64` instead of `usize`. This is not the case for FIPS.

### DIFF
--- a/mls-rs-crypto-awslc/src/x509/certificate.rs
+++ b/mls-rs-crypto-awslc/src/x509/certificate.rs
@@ -9,16 +9,16 @@ use std::{
     time::Duration,
 };
 
-use aws_lc_sys::{
+use crate::aws_lc_sys::{
     d2i_X509, i2d_X509, i2d_X509_NAME, ASN1_INTEGER_free, ASN1_INTEGER_to_BN, ASN1_TIME_free,
     ASN1_TIME_new, ASN1_TIME_set_posix, ASN1_TIME_to_posix, BN_bin2bn, BN_bn2bin, BN_free,
     BN_num_bytes, BN_to_ASN1_INTEGER, EC_KEY_get0_group, EC_KEY_get0_public_key,
     EC_POINT_point2oct, EVP_PKEY_get0_EC_KEY, EVP_PKEY_get_raw_public_key, NID_subject_alt_name,
     X509V3_set_ctx, X509_EXTENSION_dup, X509_add_ext, X509_free, X509_get0_notAfter,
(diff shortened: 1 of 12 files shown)


tomleavy (dismissed):

mls-rs-crypto-awslc/src/x509/certificate.rs
```diff
@@ -197,16 +197,17 @@ impl Certificate {
 
     pub fn public_key(&self) -> Result<SignaturePublicKey, AwsLcCryptoError> {
         unsafe {
-            let pub_key = X509_get0_pubkey(self.0);
+            let pub_key = X509_get_pubkey(self.0);
             let ec_key = EVP_PKEY_get0_EC_KEY(pub_key);
+            EVP_PKEY_free(pub_key);
```
stefunctional, line 202: Is that done too early given that `ec_key` is used after? According to [this](https://docs.openssl.org/3.0/man3/EVP_PKEY_set1_RSA/):

> EVP_PKEY_get0_EC_KEY() return the referenced key in pkey or NULL if the key is not of the correct type. The reference count of the returned key is not incremented and so the key must not be freed after use.

mulmarta: Hm, I'm not sure what this does internally but wouldn't it be crashing if we freed before use? And it's not crashing.

stefunctional: A use-after-free does not necessarily cause a crash ([example](https://godbolt.org/z/4vE4943jW)). They can lead to vulnerabilities like remote code execution.

The documentation is not very clear (there are multiple pointers involved) and I have only glanced over the C code, so I am not sure yet if there really is an issue here. Also not sure if running a tool like valgrind or some memory sanitizer while exercising this code could help. I'll try to look into it more.

mulmarta: I can also free at the end to be on the safe side. I guess it's only slightly more annoying.

mulmarta: Also there's FIPS code and non-FIPS C code which may be different, although it's unlikely.

tomleavy: get0 means no reference count increase (return an unowned reference to the internal data). Freeing would be bad here as @stefunctional points out.

stefunctional (approved):

tomleavy (approved):

PR #201, opened 2024-10-23
Fix bug where client crashes if it has a pending commit and processes a commit removing it

mulmarta:
The added test is failing in main with "attempt to subtract with overflow"

### DIFF
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -1945,9 +1945,9 @@ where
         &self,
         provisional_state: &ProvisionalState,
     ) -> Option<ProposalInfo<RemoveProposal>> {
-        match self.pending_commit {
-            Some(_) => None,
-            None => provisional_state
+        match &self.pending_commit {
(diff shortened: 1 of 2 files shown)


mls-rs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs"
-version = "0.41.4"
+version = "0.41.5"
```
tomleavy, line 3: Needs to be 0.42.0 since we merged the other CR that had breaking changes already

tomleavy (changes requested):

mls-rs/Cargo.toml
mulmarta, line 3:
```suggestion
version = "0.42.0"
```

mulmarta: Done

stefunctional (approved):

tomleavy (approved):

PR #204, opened 2024-10-28
Add more context to ms rules / identity provider callbacks

mulmarta:
Maybe useful to know more context

### Call-outs

Moving GroupContext to core means making `tree_hash` and `confirmed_transcript_hash` public (there's currently a function on ExternalGroup that returns `confirmed_transcript_hash` anyway).

### DIFF
--- a/mls-rs-core/src/identity/provider.rs
+++ b/mls-rs-core/src/identity/provider.rs
@@ -2,19 +2,12 @@
 // Copyright by contributors to this project.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-use crate::{error::IntoAnyError, extension::ExtensionList, time::MlsTime};
+use crate::{context::GroupContext, error::IntoAnyError, extension::ExtensionList, time::MlsTime};
 #[cfg(mls_build_async)]
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 
 use super::{CredentialType, SigningIdentity};
-
-#[derive(Clone, Debug, Copy)]
-pub struct CurrentEpochInfo<'a> {
-    pub group_id: &'a [u8],
-    pub epoch: u64,
-    pub extensions: &'a ExtensionList,
-}
 
 /// Identity system that can be used to validate a
 /// [`SigningIdentity`](mls-rs-core::identity::SigningIdentity)
@@ -33,7 +26,7 @@ pub trait IdentityProvider: Send + Sync {
         &self,
         signing_identity: &SigningIdentity,
         timestamp: Option<MlsTime>,
-        current_epoch: Option<CurrentEpochInfo<'_>>,
+        current_epoch: Option<&GroupContext>,
         new_extensions: Option<&ExtensionList>,
     ) -> Result<(), Self::Error>;
 
--- a/mls-rs-core/src/identity/provider.rs
+++ b/mls-rs-core/src/identity/provider.rs
@@ -2,7 +2,7 @@
 // Copyright by contributors to this project.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-use crate::{context::GroupContext, error::IntoAnyError, extension::ExtensionList, time::MlsTime};
+use crate::{error::IntoAnyError, extension::ExtensionList, group::GroupContext, time::MlsTime};
 #[cfg(mls_build_async)]
 use alloc::boxed::Box;
 use alloc::vec::Vec;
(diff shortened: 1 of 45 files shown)


tomleavy:
@mulmarta holding off on merging this for a week or so to see if we need any non-breaking changes first

mls-rs-core/src/identity/provider.rs
```diff
@@ -2,13 +2,37 @@
...
 
-use crate::{error::IntoAnyError, extension::ExtensionList, time::MlsTime};
+use crate::{context::GroupContext, error::IntoAnyError, extension::ExtensionList, time::MlsTime};
 #[cfg(mls_build_async)]
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 
 use super::{CredentialType, SigningIdentity};
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
tomleavy, line 12: Copy feels weird for this one but I guess it's ok?

mls-rs-core/src/lib.rs
```diff
@@ -9,6 +9,7 @@ extern crate alloc;
 #[cfg(all(test, target_arch = "wasm32"))]
 wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
 
+pub mod context;
```
tomleavy, line 12: We have a mod for group already, should this be a submodule of that?

mls-rs/src/group/message_processor.rs
```diff
@@ -668,17 +670,6 @@ pub(crate) trait MessageProcessor: Send + Sync {
             });
         }
 
-        let commit_effect =
```
tomleavy, line 671: Is the switch in order here so that if the update path doesn't validate we don't waste our time calculating the other stuff?

mls-rs/src/group/mod.rs
```diff
@@ -339,14 +325,35 @@ where
...
+        let context = GroupContext {
             protocol_version,
             cipher_suite,
             group_id,
+            epoch: 0,
             tree_hash,
-            group_context_extensions,
+            confirmed_transcript_hash: vec![].into(),
+            extensions: group_context_extensions,
+        };
```
tomleavy, line 336: Feels like this should be some sort of constructor function given that we need epoch and confirmed_transcript_hash to be default values?

mls-rs/src/group/proposal_filter/filtering.rs
```diff
@@ -169,10 +175,15 @@ where
         group_extensions_in_use: &ExtensionList,
         commit_time: Option<MlsTime>,
     ) -> Result<ProposalBundle, MlsError> {
+        let member_validation_context = MemberValidationContext::ForCommit {
+            current_context: self.original_context,
+            new_extensions: group_extensions_in_use,
```
tomleavy, line 180: can we change this variable name, feels like it would be the original extensions, but it isn't

mls-rs/src/group/transcript_hash.rs
```diff
@@ -19,69 +21,36 @@ use crate::{
...
+    interim_transcript_hash: &InterimTranscriptHash,
+    content: &AuthenticatedContent,
+) -> Result<ConfirmedTranscriptHash, MlsError> {
+    #[derive(Debug, MlsSize, MlsEncode)]
+    struct ConfirmedTranscriptHashInput<'a> {
+        wire_format: WireFormat,
+        content: &'a FramedContent,
+        signature: &'a MessageSignature,
     }
-}
```
tomleavy, line 37: One thing we should decide ... maybe core can have things like this in there as well. Doesn't need to change now, but it feels like it might be useful for these small crypto operations that will never change to exist in core to keep things all toegether

tomleavy (commented):
I really like the direction of this! Left some comments

mls-rs/src/group/message_processor.rs
mulmarta: That's a good reason but also we borrow `self` as immutable in `let group_state = self.group_state()` in line 626 so we can't call `self.group_state_mut()` to pop the reinit here.

mls-rs/src/group/transcript_hash.rs
mulmarta: Let's think about it. I'm not the greatest fan of the interim/confirmed transcript hash concept - one hash can be enough for security, and it could be computed more easily.

mls-rs/src/group/mod.rs
mulmarta: It only appears here, but added a constructor.

mulmarta:
I fixed the merge conflict (version in Cargo.toml), I think I addressed all the comments, and we waited 2 weeks. OK to merge @tomleavy ?

tomleavy (approved):

tomleavy:
> I fixed the merge conflict (version in Cargo.toml), I think I addressed all the comments, and we waited 2 weeks. OK to merge @tomleavy ?

Yup, I think we are good

stefunctional (approved):

PR #205, opened 2024-11-05
Improve custom proposal filtering

mulmarta:
* Basic validity checks of custom proposals must be done before they are passed to custom MlsRules - else it is too late, as they already took effect
* Local proposals are treated as by-value instead of by-reference by the proposal filter. Filtering out local proposals without filtering out their parents is error-prone.

### DIFF
--- a/mls-rs/Cargo.toml
+++ b/mls-rs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs"
-version = "0.42.0"
+version = "0.42.1"
 edition = "2021"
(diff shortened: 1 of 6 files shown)


tomleavy (approved):

stefunctional:
As discussed with Marta, this change requires some care in clients so that they don't end up in a state where they cannot commit anymore. Specifically, if a client receives a custom proposal that expands to something invalid, the client cannot commit anymore until they explicitly clear their proposal cache.

stefunctional (approved):

PR #206, opened 2024-11-05
Fix mls-rs-crypto-cryptokit not building for iOS

CaioSym:
### Issues:

TODO

### Description of changes:

Fixes  mls-rs-crypto-cryptokit to take the target aarch into account when building. Currently it always assumes macos as the target.

Explicitly added support for iphone and iphonesimulator. A more complete fix might want to consider watchOS and TVOS but I currently can't test on those.

### Call-outs:

The @available annotations SHOULD be redundant and not needed by the swift compiler since we are specifying the min target OS to be 17. However, i still get errors unless I add them. More investigation is needed on this but parking it for now since the iOS17 reqs are a hard blocker for my project to adopt crypto kit.

### Testing:

Tested that building works for the provided archs by running the following. (Assumes you've already installed the toolchains for these archs) on an M1 Mac.
```
cargo build -vv -p mls-rs-crypto-cryptokit --target aarch64-apple-ios
cargo build -vv -p mls-rs-crypto-cryptokit --target x86_64-apple-ios
cargo build -vv -p mls-rs-crypto-cryptokit --target aarch64-apple-ios-sim
cargo build -vv -p mls-rs-crypto-cryptokit --target aarch64-apple-darwin
cargo build -vv -p mls-rs-crypto-cryptokit --target x86_64-apple-darwin
```


tomleavy:
@CaioSym looks like you just need to run the fmt tool and we are good

tomleavy (approved):

mulmarta (approved):

PR #221, opened 2024-11-28
PQ KEMs behind a feature flag

mulmarta:
Implements 2 PQ replacements of DH-KEM 
* ML-KEM -- https://csrc.nist.gov/pubs/fips/203/final
* XWing -- built by combining ML-KEM with DH-KEM using a generic combiner -- https://datatracker.ietf.org/doc/draft-connolly-cfrg-xwing-kem/

## Call outs

DH-related functions **were only copied**

### Safety
I did run mls-rs-crypto-aws-lc tests with address sanitizer and stared at the code very long. Unfortunately aws-lc-rs does not currently support 
* SHAKE
* storing a decapsulation key as bytes
* deterministic key generation which is *experimental* in AWS-LC

### Testing
I would love to run some tests from the XWing RFC but unfortunately they seem to be incompatible with ML-KEM from AWS-LC and they don't contain intermediate values I could use to only test the combiner. We could generate test vectors based on the XWing implementation in the future.


tomleavy:
Looks like the entire kem combiner is uncovered code, lets fix that

mulmarta:
> Looks like the entire kem combiner is uncovered code, lets fix that

Looking. It's weird, I'm sure I tested it...

tomleavy (dismissed):

mls-rs-crypto-awslc/src/kem/ml_kem.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    let mut secret_key = vec![0u8; secret_key_len];
+    let mut public_key = vec![0u8; public_key_len];
+
+    let err = 1 != EVP_PKEY_get_raw_private_key(pkey, secret_key.as_mut_ptr(), &mut secret_key_len)
+        || 1 != EVP_PKEY_get_raw_public_key(pkey, public_key.as_mut_ptr(), &mut public_key_len);
+
+    EVP_PKEY_CTX_free(ctx);
+    EVP_PKEY_free(pkey);
```
stefunctional, line 209: Should the key be freed before the context? The key was created using the context and I don't know if the context is used when freeing the key.

mls-rs-crypto-rustcrypto/src/ecdh.rs
```diff
@@ -97,6 +97,10 @@ impl DhType for Ecdh {
     fn secret_key_size(&self) -> usize {
         self.0.secret_key_size()
     }
+
+    fn public_key_size(&self) -> usize {
+        self.0.secret_key_size()
```
stefunctional, line 102: Should the call be forwarded to `public_key_size` instead?
```suggestion
        self.0.public_key_size()
```

mls-rs-crypto-webcrypto/src/ec/ecdh.rs
```diff
@@ -165,6 +165,10 @@ impl DhType for Ecdh {
     fn public_key_validate(&self, _key: &HpkePublicKey) -> Result<(), Self::Error> {
         Ok(())
     }
+
+    fn public_key_size(&self) -> usize {
+        self.curve.secret_key_size()
```
stefunctional, line 170: Is that right or should it be as below?
```suggestion
        self.curve.public_key_size()
```

stefunctional (commented):
Just a couple of questions.

mls-rs-crypto-awslc/src/kem/ml_kem.rs
mulmarta: I did run it with sanitizer - wouldn't it catch problems with that?

So the doc for EVP_PKEY_keygen says "it sets |*out_pkey| to a newly-allocated |EVP_PKEY|" which I understood as "allocates independent key". I also found a couple of examples that free context and key [here](https://github.com/aws/aws-lc/blob/5084e19727f5c922a25361bb639b73d266cd8948/util/benchmark/ec/p256_awslc_ossl/src/benchmark_ecdh.c#L171) and [here](https://github.com/aws/aws-lc/blob/5084e19727f5c922a25361bb639b73d266cd8948/crypto/fipsmodule/kem/README.md?plain=1#L105)

I can also ask the guy who implemented this for us - do you think I should?

stefunctional: The examples you linked both seem to free the key before the context, which is the opposite order of what is done in this PR. I don't know if sanitizer would catch such an issue. FWIW I had a quick look at the C code and I didn't see any context stored in the key or used when freeing it, so maybe the destruction order does not matter?

Asking the implementer would remove any doubt, but is it worth it? I would just follow the order in the examples you linked to be safe, but I am not opposed to merging this as-is if you think that is not worth the trouble -- your extensive testing didn't reveal any issue.

stefunctional (approved):
Thank you!

tomleavy (approved):

PR #224, opened 2024-12-11
feat(mls-rs): Verify the update path even in case of a self removal

ManevilleF:
### Description of changes:

In the case of a self removal commit we early return before verifying the `update_path`, This PR drops the early return and self removals verify the `update_path` and adds some checks in following operations to avoid:
- Applying the `update_path` for a self removal
- Updating the key schedule for a self removal

### DIFF
--- a/mls-rs/src/group/message_processor.rs
+++ b/mls-rs/src/group/message_processor.rs
@@ -656,19 +656,8 @@ pub(crate) trait MessageProcessor: Send + Sync {
             return Err(MlsError::CommitMissingPath);
         }
 
-        if let Some(remove_proposal) = self.removal_proposal(&provisional_state) {
-            let new_epoch = NewEpoch::new(self.group_state().clone(), &provisional_state);
-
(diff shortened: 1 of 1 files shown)


mulmarta:
I'm wondering @ManevilleF how much security this check is adding? We're verifying a path that we're going to immediately delete since we're removed and its not useful for us. And note that if we're removed, we anyway can't verify _everything_ inside the commit such as the confirmation tag.

mulmarta (approved):

ManevilleF:
> So if the committer is honest, it won't generate a bad path (if it's a ghost, commit will fail). If it's dishonest, it can fragment anyway. So the space of fragmenting attacks this prevents seems rather narrow?

fragmentation by malicious committers is unavoidable, yet. This just closes one small door even though as you pointed out there's a blatant gateway opened to cause fragmentation.
This could also impact ill-implemented clients (committer commits even though its credential is revoked) so I think it's still worth it

tomleavy (approved):

PR #229, opened 2024-12-23
Use stored state for apply-pending-commit

mulmarta:
* Implement mls-codec for bool
* Make apply-pending-commit just write new state


mls-rs/src/group/message_processor.rs
```diff
@@ -140,11 +140,60 @@ pub enum CommitEffect {
...
+            2u8 => Ok(Self::Removed {
+                new_epoch: NewEpoch::mls_decode(reader)?.into(),
+                remover: Sender::mls_decode(reader)?,
+            }),
+            3u8 => Ok(Self::ReInit(ProposalInfo::mls_decode(reader)?)),
+            _ => Err(mls_rs_codec::Error::UnsupportedEnumDiscriminant),
+        }
+    }
+}
+
```
tomleavy, line 196: We do have at least some support for enum with associated data in the macro. Was there something in particular that was a problem?

mls-rs/src/group/snapshot.rs
```diff
@@ -39,7 +41,8 @@ pub(crate) struct Snapshot {
     key_schedule: KeySchedule,
     #[cfg(feature = "by_ref_proposal")]
     pending_updates: SmallMap<HpkePublicKey, (HpkeSecretKey, Option<SignatureSecretKey>)>,
-    pending_commit: Option<CommitGeneration>,
+    #[mls_codec(with = "mls_rs_codec::byte_vec")]
+    pending_commit: Vec<u8>,
     signer: SignatureSecretKey,
```
tomleavy, line 46: Doesn't this break a user who has already exported a snapshot? I think you need v2 and a custom MlsDecode to handle it?

mls-rs/src/group/message_processor.rs
mulmarta: Mostly the problem was not supporting named fields --- `Removed { new_epoch }` --- I didn't want to blow up this PR by implementing that, I could create an issue

mls-rs/src/group/snapshot.rs
mulmarta: Note that the old CommitGeneration is not usable anymore, so the pending commit will be lost.

tomleavy (dismissed):

mls-rs/src/group/message_processor.rs
tomleavy: Ok yeah lets make an issue for it, I think this can be solved

tomleavy (approved):

mls-rs/src/group/message_processor.rs
mulmarta: #232

mls-rs/src/group/snapshot.rs
```diff
@@ -159,24 +207,24 @@ where
...
-        let mut snapshot = self.snapshot();
+        let mut snapshot = self.snapshot()?;
         snapshot.state.public_tree.nodes = Default::default();
 
         self.state_repo.write_to_storage(snapshot).await
     }
 
-    pub(crate) fn snapshot(&self) -> Snapshot {
-        Snapshot {
+    pub(crate) fn snapshot(&self) -> Result<Snapshot, MlsError> {
```
stefunctional, line 216: Why was it changed to return a `Result`? It seems `Ok` is always returned.

stefunctional (approved):

mls-rs/src/group/snapshot.rs
mulmarta: Good catch, I will fix in a followup

PR #231, opened 2024-12-26
Work around rust < 1.78 crash

glandium:
Somehow the DWARF info generated by the compiler for the `hash`-replacement assignment is confusing to LLVM, which crashes.

By using a different form for the same operation, the compiler is happy.


mulmarta (approved):

khuey:
It's safe to revert this once your MSRV is 1.87.0 or higher. See rust-lang/rust#138818

PR #236, opened 2025-01-12
Remove overloaded public methods from X509IdentityProvider

mgeisler:
### Description of changes:

While using a `X509IdentityProvider` in a generic context, I discovered that it has a few methods which I found confusing:

- `X509IdentityProvider::identity` takes 1 argument, but `IdentityProvider::identity` takes 2 arguments!
- `X509IdentityProvider::valid_successor` takes 2 arguments, but `IdentityProvider::valid_successor` takes 3 arguments!
                                                         
I ended up solving this with the fully-qualified `<Type as Trait>::method` syntax, but I think the situation can be avoided all together by just implementing the trait directly.                                                              

Please let me know what you think of this — I think it's simpler without the "double" implementations, but perhaps I'm missing something.

### Call-outs:

This is a SemVer breaking change, so I've bumped the version number of mls-rs-identity-x509. Since the type is re-exported by mls-rs, I bumped its version number as well.

### Testing:

Tests we updated to pass in the necessary parameters, even though we know they are not used in the tests.


mgeisler:
The "[Native / LintAndFormatting (pull_request)](https://github.com/awslabs/mls-rs/actions/runs/12744195695/job/35515632618?pr=236)" failure seems unrelated — it's about an unexpected lint, which became a warning recently.

mulmarta:
I fixed the warnings in #237 - we only fixed them on 1.x-main and forgot main

mulmarta (approved):

mgeisler:
> I fixed the warnings in #237 - we only fixed them on 1.x-main and forgot main

Thanks for merging the fix back! I should have done that, but got distracted today :smile:

tomleavy (approved):

mgeisler:
By the way, do you plan on _merging_ the `1.x-main` branch into `main` later? I'm asking because I wonder how you will ensure changes like this make it to the 1.x release?

tomleavy:
Yeah that's the idea. Once we have enough for alpha 1 we will merge and create a branch off right before that for bug fixes on 0.x

Issue #211, opened 2024-11-11
[1.x] Externalize storage


tomleavy:
### Background:
As part of our 1.x goals, we want to externalize all storage objects that interact with mls-rs so internal functionality is not directly dependent on a user-provided storage mechanism.

### Key Package Storage
Currently building a client requires that you provide an implementation of the KeyPackageStorage trait that we define in mls-rs-core. A Client uses this to maintain storage on your behalf when certain operations take place:

When a key package is generated with [Client::generate_key_package_message](https://docs.rs/mls-rs/latest/mls_rs/client/struct.Client.html#method.generate_key_package_message), the Client will store the associated private key in the KeyPackageStorage implementation

When a group is joined with [Client::join_group](https://docs.rs/mls-rs/latest/mls_rs/client/struct.Client.html#method.join_group), the Client will first retrieve the private key that corresponds with the key package that was used to add it to the group and also delete that private key when the resulting Group is saved.

- [ ] #207 
- [ ] #209 

### PSK Storage
PSK storage is currently set when building a client and is consumed whenever a PSK is required:

* Joining a group with [Client::join_group](https://docs.rs/mls-rs/latest/mls_rs/client/struct.Client.html#method.join_group)
* Processing a commit that requires knowledge of a PSK with [Group::process_incoming_message](https://docs.rs/mls-rs/latest/mls_rs/group/struct.Group.html#method.process_incoming_message)
* Producing a commit introduces or requires knowledge of a PSK with [Group::CommitBuilder](https://docs.rs/mls-rs/latest/mls_rs/group/struct.CommitBuilder.html)

mulmarta:
PSK is also used in `Client::external_commit_builder`

PR #238, opened 2025-01-14
Remove psk store

tomleavy:
### Issues:

Work towards #211 

### Description of changes:

Remove the PreSharedKeyStorage trait from Client and Group operations

### Call-outs:

There are some obvious places where there can be optimizations like removing clones or passing an iterator instead of a Vec. Since there are other changes in those areas I am keeping this PR limited to just getting things working and we can take care of those other items later.

### Testing:

Adapted all existing tests


mls-rs/src/group/commit/processor.rs
```diff
@@ -282,6 +325,16 @@ impl<C: ClientConfig> CommitProcessor<'_, C> {
...
+
+    pub fn required_psks(&self) -> Vec<ExternalPskId> {
+        self.0
+            .proposals
+            .psk_proposals()
+            .iter()
+            .filter_map(|p| p.proposal.external_psk_id())
+            .cloned()
+            .collect()
+    }
```
mulmarta, line 337: What about resumption PSKs? They're handled explicitly in other places which I think makes sense

mls-rs/src/group_joiner.rs
```diff
@@ -337,39 +372,54 @@ mod tests {
...
             .build()
             .await
             .unwrap();
 
         let joiner = alice
             .group_joiner(&commit.welcome_messages[0], kp_alice.key_package_data)
-            .await
-            .unwrap();
+            .unwrap()
+            .await;
```
mulmarta, line 399: how is the sequence `unwrap().await` possible here?

mls-rs/test_harness_integration/src/main.rs
```diff
@@ -734,6 +743,7 @@ impl MlsClientImpl {
 
         let roster = group.roster().members();
 
+        let group_clone = group.clone();
```
mulmarta, line 746: This clone seems useless?

tomleavy: Unfortunately it is required because once you create a builder or processor that references `&mut Group` then you can no longer call any functions on Group until you complete the building process. It's actually a problem I hit in a few places within tests, and I think we will have to separate out the structures in terms of parsing vs the actual building so that the &mut reference is not an issue.

mls-rs/src/group_joiner.rs
tomleavy: fixed

mls-rs/src/group/commit/processor.rs
tomleavy: fixed

mls-rs/src/group/external_commit.rs
```diff
@@ -168,20 +137,9 @@ impl<C: ClientConfig> ExternalCommitBuilder<C> {
...
-
-        // We store proposal to verify authenticity later. At this point this may not be possible if we
-        // don't have the tree.
-        self.received_custom_proposals.push(plaintext);
-
-        Ok(self.with_proposal(*proposal))
+    pub fn with_received_custom_proposal(mut self, proposal: MlsMessage) -> Self {
+        self.received_custom_proposals.push(proposal);
+        self
     }
```
mulmarta, line 143: How does this work with custom proposal rules being applied outside? I forgot to add a function for this but it's like in CommitBuilder:
```handle_custom_proposals(commit_builder.proposals_mut());  // expand custom proposals, add more proposals if needed```

mulmarta (approved):

PR #247, opened 2025-01-23
Simplify X.509 certificate validity check

mgeisler:
### Description of changes:

Refactor of the `verify_time` code for X.509 certificates: instead of chaining `Option` values, I used a simple `if` statement. I find that clearer.

### DIFF
--- a/mls-rs-crypto-rustcrypto/src/x509/validator.rs
+++ b/mls-rs-crypto-rustcrypto/src/x509/validator.rs
@@ -142,9 +142,11 @@ fn verify_time(cert: &Certificate, time: MlsTime) -> Result<(), X509Error> {
     let not_before = validity.not_before.to_unix_duration().as_secs();
     let not_after = validity.not_after.to_unix_duration().as_secs();
 
-    (not_before <= now && now <= not_after)
-        .then_some(())
-        .ok_or_else(|| X509Error::ValidityError(now, format!("{cert:?}")))
+    if not_before <= now && now <= not_after {
+        Ok(())
+    } else {
+        Err(X509Error::ValidityError(now, format!("{cert:?}")))
+    }
 }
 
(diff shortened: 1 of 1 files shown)


mgeisler:
I found this code while looking at validating X.509 certificates. I would also like to discuss the `X509Error::ValidityError(now, format!("{cert:?}"))` error returned with a big debug print of the certificate :smile: It doesn't seem super useful?

Would there be appetite for changing `X509Error::ValidityError` to have fields with:

* A certitificate serial number (if that's a useful identifier?)
* The `now` time passed in
* The `notBefore` time
* The `notAfter` time

I will need to parse these things myself anyway, so changing this is not strictly necessary for me — but it could perhaps be a nice little cleanup.

mulmarta:
Yeah definitely having the very large cred in the debug print is suboptimal. I guess time makes sense, as the most likely reason for failing.

mgeisler:
> Yeah definitely having the very large cred in the debug print is suboptimal. I guess time makes sense, as the most likely reason for failing.

From what I see, it's the only place this enum variant is constructed, so it should be fine to adjust.

mgeisler:
Ups, I fixed the formatting mistake... I've somehow managed to disable `rustfmt` on my work computer :man_facepalming:

mulmarta (approved):

tomleavy (approved):

tomleavy:
> > Yeah definitely having the very large cred in the debug print is suboptimal. I guess time makes sense, as the most likely reason for failing.
> 
> From what I see, it's the only place this enum variant is constructed, so it should be fine to adjust.

Merging this one, feel free to put up another one to change this

mgeisler:
> Merging this one, feel free to put up another one to change this

Thank you, that was what I was hoping.

PR #248, opened 2025-01-24
Group API to encrypt/decrypt using the leaf-node HPKE keys

emarteca:
### Description of changes:

Hi! I would like to add the ability for the current member of an mls-rs Group to encrypt a message to another member in the Group, using the recipient's HPKE public key. Symmetrically, I would like the ability for the current member of a group to be able to decrypt messages encrypted to them with this new HPKE encrypt function, which requires its HPKE private key. These keys are not exposed through a public API, and this remains true with my proposed changes.

In this PR, I've implemented this by adding two public methods to the mls-rs Group. Under the hood, these methods use the HPKE encrypt/decrypt functionality available in the CipherSuiteProvider through `hpke_seal` and `hpke_open`.

These methods have the following signatures:
```
pub fn hpke_encrypt_to_recipient(
        &self,
        recipient_index: u32,
        context_info: &[u8],
        associated_data: Option<&[u8]>,
        plaintext: &[u8],
    ) -> Result<HpkeCiphertext, MlsError>;

pub fn hpke_decrypt_for_current_member(
        &self,
        context_info: &[u8],
        associated_data: Option<&[u8]>,
        hpke_ciphertext: HpkeCiphertext,
    ) -> Result<Vec<u8>, MlsError>;
```

They take all of the arguments required to call the `hpke_seal` and `hpke_open` respectively.

### Testing:
I've added some unit tests to `src/group/mod.rs`. These test the following scenarios:
- simple HPKE encrypt/decrypt
- non-recipient member in the group cannot HPKE decrypt
- recipient can HPKE decrypt a ciphertext even if the group has changed since it was encrypted


mulmarta (changes requested):
Thanks @emarteca ! That's a nice feature to have for sure. But I think we should be using the MLS extension called [targeted messages](https://messaginglayersecurity.rocks/mls-extensions/draft-ietf-mls-extensions.html#name-targeted-messages) that does what you need but more securely.

Basically we need more domain separation, in order to make sure that no matter how an application uses this feature, it cannot be used to break security of the main MLS protocol. The way it currently is, an HPKE ciphertext generated by MLS can be decrypted using `hpke_decrypt_for_current_member`. 

So for example, say there is an application where recipients of messages sent this way, call them targeted messages, regularly publish them on a blockchain without parsing. Now an adversary can take an HPKE ciphertext encrypted by MLS to Alice's leaf, deliver it to Alice disguised as a targeted message, read the plaintext from the blockchain. The plaintext is a path secret that allows to decrypt MLS messages.

mulmarta:
Hey @emarteca using [SafeEncryptWithContext](https://www.ietf.org/archive/id/draft-barnes-mls-appsync-01.html#name-hybrid-public-key-encryptio) is also safe and more flexible than targeted messages. I think you can add new functions that implement SafeEncryptWithContext (and decrypt) by calling your new functions black-box?

Then we can create an issue to implement targeted messages using SafeEncryptWithContext.

emarteca:
@mulmarta thanks for the discussion and direction with this!

I've updated this PR to have both the original API but gated behind a flag, and to add the safe encrypt/decrypt with context as described in [this doc you linked](https://github.com/bifurcation/mls-appsync/blob/app-api/draft-barnes-mls-appsync.md#component-ids). I've made some minor changes to the API in the doc, namely:
* addition of an associated data parameter (as discussed offline)
* rather than having these APIs take in the public/private keys directly, I set them up similarly to my original APIs, so that the encrypt function takes a recipient index (and the public key for that member is retrieved from its leaf node), and the decrypt function is for the current group member (and uses its own leaf node private key).

I copied over the same tests as I had for the original API and adapted them to the new safe encrypt/decrypt with context.

Force pushed to rebase onto new main head.

Please let me know what you think and if you'd like any changes, thank you!

mulmarta:
Sorry about all the different build options that we support, that is a bit hard to maintain. To fix non-std build you have to explicitly import `alloc::vec::Vec` and such. To fix the async build you need to make everything async if `mls_build_async` build flag is set, see [example](https://github.com/awslabs/mls-rs/blob/f7a8f8757a261ae8ec180c5cf43125d1a28c7506/mls-rs/src/group/util.rs#L34).

mls-rs/src/group/component_operation.rs
```diff
@@ -0,0 +1,32 @@
...
+    const ENCRYPT_LABEL: &'static str = "MLS 1.0 Application";
+
+    fn from_bytes(bytes: Vec<u8>) -> Result<Self, MlsError> {
+        Self::mls_decode(&mut bytes.as_slice()).map_err(Into::into)
+    }
+
+    fn get_bytes(&self) -> Result<Vec<u8>, MlsError> {
+        self.mls_encode_to_vec().map_err(Into::into)
+    }
+}
```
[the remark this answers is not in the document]
emarteca: Got it -- I misunderstood how the label was meant to be represented, and thought it was supposed to be ENCRYPT_LABEL. Fixed it.

mls-rs/src/group/component_operation.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::tree_kem::hpke_encryption::HpkeEncryptable;
+use mls_rs_codec::{MlsDecode, MlsEncode, MlsSize};
+
+pub type ComponentID = u32;
+
+#[derive(Clone, Debug, PartialEq, MlsSize, MlsEncode, MlsDecode)]
+pub struct ComponentOperationLabel {
+    component_id: ComponentID,
+    context: Vec<u8>,
+}
```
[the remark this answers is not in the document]
emarteca: Fixed it (same issue as above)

emarteca:
No problem, thanks for explaining and for the example! Hopefully the build works now -- I tested locally but I can't see how to run the entire CI pipeline.

mulmarta:
> No problem, thanks for explaining and for the example! Hopefully the build works now -- I tested locally but I can't see how to run the entire CI pipeline.

I don't know how that works, but it's possible to run the "push" actions on a fork, e.g. this fork runs actions https://github.com/bifurcation/mls-rs/actions

emarteca:
Thank you :pray: I _think_ I got it this time. I got the CI running in a different branch in my fork of the repo and it's passing there.

mls-rs/src/group/mod.rs
```diff
@@ -602,6 +604,155 @@ where
...
+    /// Takes `context_info`, `associated_data`, and `plaintext`.
+    /// Returns `ciphertext` and `kem_output` inside `HpkeCiphertext`.
+    ///
+    /// WARNING: The message sender is not authenticated.
+    #[cfg(all(feature = "non_domain_separated_hpke_encrypt_decrypt", feature = "ffi"))]
+    #[cfg_attr(
+        not(mls_build_async),
+        maybe_async::must_be_sync,
+        safer_ffi_gen::safer_ffi_gen_ignore
+    )]
```
mulmarta, line 639: This is horrible on our part. FWIW we hope to get rid of safer-ffi and rely instead on mls-rs wrappers using uniffi.

mls-rs/src/group/mod.rs
```diff
@@ -602,6 +604,155 @@ where
...
+    pub async fn safe_decrypt_with_context_for_current_member(
+        &self,
+        component_id: ComponentID,
+        context: &[u8],
+        associated_data: Option<&[u8]>,
+        hpke_ciphertext: HpkeCiphertext,
+    ) -> Result<Vec<u8>, MlsError> {
+        let component_operation_label =
+            ComponentOperationLabel::new(component_id, context.to_vec());
+        self.hpke_decrypt_for_current_member_with_generic_context(
```
mulmarta, line 748: To make it a bit more efficient, I think that you can avoid the `to_vec` call if you define `ComponentOperationLabel` as follows (I don't think `MlsDecode` is needed for anything)
```
#[derive(Clone, Debug, PartialEq, MlsSize, MlsEncode)]
pub struct ComponentOperationLabel<'a> {
    label: &'static [u8],
    component_id: ComponentID,
    context: 'a[u8],
}
```

mulmarta (approved):
Thanks a lot!

I left a comment about a small efficiency improvement, but otherwise looks good.

mls-rs/src/group/mod.rs
emarteca: Good idea; done!

mls-rs/src/group/mod.rs
emarteca: No problem! It was easy enough to figure out the required cfgs. But that'll be nice in the future :smiley:

mulmarta (dismissed):

emarteca:
Thanks for your all your help with this!

mls-rs/Cargo.toml
```diff
@@ -30,6 +30,7 @@ out_of_order = ["private_message"]
 prior_epoch = []
 by_ref_proposal = []
 psk = []
+non_domain_separated_hpke_encrypt_decrypt = []
 x509 = ["mls-rs-core/x509", "dep:mls-rs-identity-x509"]
 rfc_compliant = ["private_message", "custom_proposal", "out_of_order", "psk", "x509", "prior_epoch", "by_ref_proposal", "mls-rs-core/rfc_compliant"]
 last_resort_key_package_ext = ["mls-rs-core/last_resort_key_package_ext"]
```
mulmarta, line 36: You need to bump the version @emarteca else I can't publish it.

stefunctional (dismissed):

mls-rs/Cargo.toml
emarteca: Ah, thank you!

emarteca:
Force-push to rebase on new main head.

Thanks for the approvals! Looks like they are expired b/c of the version update + rebase :sweat_smile:

stefunctional (approved):

mulmarta (approved):

PR #190, opened 2024-09-05
Add P384 suport for RustCrypto

mulmarta:
### Description of changes:

RustCrypto now supports P384.

### Testing:

Covered by the core test.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT license.

PR #251, opened 2025-01-31
Add support for P384 public key algorithm

mgeisler:
### Description of changes:

I found that I could not use root certificates which use a P384 public key algorithm. More specifically, passing in a certificate with

```
$ openssl x509 -noout -text -in root.der
…
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (384 bit)
                pub:
                    …
                ASN1 OID: secp384r1
                NIST CURVE: P-384
…
```

gives me

```
EcX509Error(UnsupportedPublicKeyAlgorithm("ObjectIdentifier(1.2.840.10045.2.1)"))
```

back. The changes here seem to fix this, but I’ll admit that they were made very mechanically based on the existing code.

This would be a continuation of the work in #190.


mgeisler:
@mulmarta, I hope the changes here make sense :smile: I believe they finish the work you started in #190 — but I might be missing something since I'm not an expert on X.509 certificates and their fields.

mulmarta:
> @mulmarta, I hope the changes here make sense 😄 I believe they finish the work you started in #190 — but I might be missing something since I'm not an expert on X.509 certificates and their fields.

Oh I totally forgot about X.509. I keep forgetting, since it's not useful for me (and too complicated for its own good). Do you think it would make sense to add a test?

mulmarta (dismissed):

mgeisler:
> Do you think it would make sense to add a test?

I would love to, I just haven't looked up the commands to generate the right certificate 🙂

mgeisler:
Hi @mulmarta, I tried generating a certificate with certificatetools.com, which I believe was also used for the others.

I generated a self-signed CA certificate, but when I try loading it, I'm told that the signature is invalid:

```
running 1 test
test x509::validator::tests::can_validate_p384_cert ... FAILED

failures:

---- x509::validator::tests::can_validate_p384_cert stdout ----
thread 'x509::validator::tests::can_validate_p384_cert' panicked at mls-rs-crypto-rustcrypto/src/x509/validator.rs:290:91:
called `Result::unwrap()` on an `Err` value: EcSignerError(InvalidSignature)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The test failure is in the second commit on the branch.

Do you have any tips for debugging this kind of error?

Do you know how to generate a certificate in some other way? I can try running the OpenSSL commands the site spits out, but I assume it'll give me the same kind of error.

Any help would be much appreciated!

mgeisler:
> Had a quick look at your cert @mgeisler , you have the wrong hash.

Ah, amazing! I'll try updating it!

mulmarta (dismissed):

mgeisler:
I fixed the formatting, sorry about that!

stefunctional (approved):

mulmarta (approved):

PR #252, opened 2025-02-05
Bump versions of everything since wasm-bindgen version is incompatible

mulmarta:

### DIFF
--- a/mls-rs-crypto-openssl/Cargo.toml
+++ b/mls-rs-crypto-openssl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-openssl"
-version = "0.12.1"
+version = "0.14.0"
 edition = "2021"
 description = "OpenSSL based CryptoProvider for mls-rs"
 homepage = "https://github.com/awslabs/mls-rs"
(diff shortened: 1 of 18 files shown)


stefunctional:
Only the patch version number is incremented meaning these new versions are considered compatible with the previous ones as far as cargo is concerned. Is that intended?

mls-rs-crypto-openssl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-crypto-openssl"
-version = "0.12.0"
+version = "0.14.0"
```
stefunctional, line 3: Out of curiosity, is there a reason 0.13.0 was skipped? Even if inadvertence is the answer, that's fine and you don't have to go through the pain of updating everything all over again.

stefunctional (approved):

mls-rs-crypto-openssl/Cargo.toml
mulmarta: Not completely inadvertence, the compiler got stuck somehow at some point, failing to catch the update to 13. Let's skip 13.

PR #254, opened 2025-02-14
Cleaning up `non_domain_separated_hpke_encrypt_decrypt` cfgs

emarteca:
Cleaning up the cfgs so users can use non_domain_separated_hpke_encrypt_decrypt without ffi feature enabled

In the current state, the `cfg`s for the `non_domain_separated_hpke_encrypt_decrypt` flagged-out functions are configured such that they can only be used if a user has the `ffi` feature also enabled. This requires a dependency on `safer_ffi_gen`. (This is my mistake, for how I configured these)

This PR just changes the configuration to remove this dependency, so users can use the `non_domain_separated_hpke_encrypt_decrypt` functions even without `ffi`.

### DIFF
--- a/mls-rs/Cargo.toml
+++ b/mls-rs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs"
-version = "0.45.0"
+version = "0.45.1"
 edition = "2021"
 description = "An implementation of Messaging Layer Security (RFC 9420)"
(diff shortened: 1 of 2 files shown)


mulmarta (approved):
FWIW you can just move functions out of the FFI impl block, then you don't care about "ffi" features

mulmarta (approved):

tomleavy (approved):

PR #255, opened 2025-02-18
Add message description function from 1.x

mulmarta:
Knowing content type of a message is very useful.

### DIFF
--- a/mls-rs/src/group/framing.rs
+++ b/mls-rs/src/group/framing.rs
@@ -342,6 +342,45 @@ impl From<&PrivateMessage> for PrivateContentAAD {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum MlsMessageDescription<'a> {
+    Welcome {
+        key_package_refs: Vec<&'a KeyPackageRef>,
+        cipher_suite: CipherSuite,
+    },
+    ProtocolMessage {
+        group_id: &'a [u8],
+        epoch_id: u64,
+        content_type: ContentType, // commit, proposal, or application
+    },
+    GroupInfo,
+    KeyPackage,
+}
+
+impl MlsMessage {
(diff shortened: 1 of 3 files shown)


mls-rs/src/group/framing.rs
```diff
@@ -342,6 +342,45 @@ impl From<&PrivateMessage> for PrivateContentAAD {
...
+                content_type: p.content.content_type(),
+            },
+            #[cfg(feature = "private_message")]
+            MlsMessagePayload::Cipher(c) => MlsMessageDescription::ProtocolMessage {
+                group_id: &c.group_id,
+                epoch_id: c.epoch,
+                content_type: c.content_type,
+            },
+            MlsMessagePayload::GroupInfo(_) => MlsMessageDescription::GroupInfo,
+            MlsMessagePayload::KeyPackage(_) => MlsMessageDescription::KeyPackage,
```
tomleavy, line 379: Shouldn't this give you access to the underlying key package data or group info data? I know that in 1.x we focused on what we need inside execution, but just for completeness it would be worth having more stuff here?

mulmarta: It's not checking the signature though. I'm not sure it's safe to use?

tomleavy: I think so, it's just a description, you can't really use what it tells you for anything bad because other API won't accept it

mulmarta: You can't join a group or add somebody using it, but you may make decisions based on say custom group context extensions in group info. But the authenticity of that is not validated. I'm worried that this is a footgun for applications

tomleavy (approved):

stefunctional (approved):

tomleavy (approved):

PR #262, opened 2025-03-13
Implement Padme padding mode.

cindylindeed:
See Algorithm 1 in https://www.petsymposium.org/2019/files/papers/issue4/popets-2019-0056.pdf.

### Description of changes:

Add new Padme padding mode for private messages.

### Testing:

* Added test vectors calculated using reference implementation found @ https://lbarman.ch/blog/padme/#implementation.
* Added Padme to all existing padding tests.


cindylindeed:
Removed the u32::MAX test since it overflows on 32-bit machines. Double-checked that the existing StepFunction padding scheme also overflows and panics on those values (as expected).

tomleavy:
Thanks for submitting this, seems like a great addition. @jalwen did the step function, he would be best to take a look.

mulmarta:
Thanks! I will have a look at the function.

@cindylindeed you can run `cargo clippy --fix`  tool to automatically fix clippy warnings.

tomleavy (approved):

tomleavy (approved):

mls-rs/src/group/padding.rs
```diff
@@ -31,6 +35,24 @@ impl PaddingMode {
...
+                // E <- floor(log2(L))
+                // S <- floor(log2(E)) + 1
+                // z <- E - S
+                // m <- (1 << z) - 1
+                // len' <- (L + m) & ~m
+
+                let e: u32 = content_size.ilog2(); // l’s floating-point exponent
+                let s: u32 = e.ilog2() + 1; // number of bits to represent e
+                let num_zero_bits: u32 = e - s; // number of low bits to set to 0
+                let bitmask: usize = (1 << num_zero_bits) - 1; // create a bitmask of 1s
```
mulmarta, line 53:
```suggestion
                let num_one_bits: u32 = e - s; // number of low bits to set to 1
                let bitmask: usize = (1 << num_one_bits) - 1; // create a bitmask of 1s
```
Maybe I'm confused, but we're fixing to 1's to increase the length?

mulmarta (approved):

mls-rs/src/group/padding.rs
mulmarta: Can be a follow-up if needed/

cindylindeed: Ye, the usage here is a little convoluted wrt the variable naming. Note that `bitmask` is ultimately used as `!bitmask`. Instead of setting `E = exponent` number of low bits to 0 (which is equivalent to padding to powers of 2), we set `E-S` low bits to 0 (which allows more granular padding, though more leakage of size). Before setting `E-S` low bits to 0, we add `E-S` to the value to round up if needed.

We could rename some variables to make this clearer, though it'll then diverge from the reference pseudocode in the paper. LMK what you prefer.

PR #263, opened 2025-03-15
API to validate `CustomProposal` from prior epochs

emarteca:
This PR adds a new API, `validate_custom_proposal` , on `Group`.  This validates the membership tag and signature of a custom proposal message from the current or a saved `PriorEpoch`.
This new API function:
* takes an `MlsMessage` 
* validates its membership tag and signature using the existing function `verify_plaintext_authentication`
* asserts that the message must be a `CustomProposal`, and 
* returns the `authenticated_data` and `Sender` associated with the proposal. 

It uses a more general function we added as a helper, `validate_public_message` which validates any public message (and returns the `AuthenticatedContent`).

In order to allow `verify_plaintext_authentication` to validate messages from prior epochs, we added a `membership_tag` field to `PriorEpoch`, which stores the KeyScheduler's current membership tag when the epoch is stored. Then, we modified `verify_plaintext_authentication` to closely match the signature of `verify_auth_content_signature`, to allow information from prior epochs (group context, and signature public keys container) to be passed through (since `PriorEpoch` does not store the entire group state). 

The current functionality of `process_incoming_message` is unmodified. Only public messages from the current epoch can be applied to the group. 

In our new API, we add support for looking up prior epochs when validating the public messages, following the strategy in `decrypt_incoming_ciphertext` which adds support for `PriorEpoch` validation for private messages.

----

All of this functionality is gated behind a new feature flag `validate_public_message_prior_epoch`. 
This PR also adds tests for the new API and the helper function.


mls-rs/src/group/epoch.rs
```diff
@@ -30,6 +30,8 @@ pub(crate) struct PriorEpoch {
     pub(crate) self_index: LeafIndex,
     pub(crate) secrets: EpochSecrets,
     pub(crate) signature_public_keys: Vec<Option<SignaturePublicKey>>,
+    #[cfg(feature = "validate_public_message_prior_epoch")]
```
tomleavy, line 33: I'm concerned about the format of stored data being different when the feature is on vs off. I can discuss with @mulmarta, but one suggestion is that we can add a feature to the mls encoding macro to have fields at the end of a struct be optional and automatically filled in with the Default value for that type. 

The end result is something like  this:

```
#[mls(default)]
membership_key: Option<Vec<u8>>
```

mulmarta: Yeah, I had the same thoughts. MLS encoding is not great for these stored values. It would be better to have something more flexible.

mulmarta (commented):
Switching on this feature decreases MLS security because old membership keys are stored. I think there should be a large WARNING somewhere in a comment (ideally on the feature somehow) explaining security consequences of this.

It would also be nice to explain the possible use cases for this, since receiving old proposals seems not very useful for MLS - the only purpose of proposal is to be committed to. Possibly a link to some protocol using this?

cindylindeed:
> Switching on this feature decreases MLS security because old membership keys are stored. I think there should be a large WARNING somewhere in a comment (ideally on the feature somehow) explaining security consequences of this.
> 
> It would also be nice to explain the possible use cases for this, since receiving old proposals seems not very useful for MLS - the only purpose of proposal is to be committed to. Possibly a link to some protocol using this?

@emarteca - Please add the following as a comment:

WARNING: This API is not recommend and is unsafe for regular usage as it reduces the guarantees of RFC9420. Since prior epochs' membership keys are stored, an attacker that compromises a client is able to send valid MLS PublicMessages appearing to originate from the compromised client from prior epochs.

This API is used by the GSMA's Rich Communication Suite for certain types of messages. See section 7.6 for specifics https://www.gsma.com/solutions-and-impact/technologies/networks/wp-content/uploads/2025/03/RCC.16-v1.0.pdf.

mls-rs/src/group/epoch.rs
emarteca: This makes sense. Is it something that we could do in a followup PR, since it sounds like some changes would be required to the MlsEncode trait?

emarteca:
Thanks for your comments! I've updated the changes to include the warning @cindylindeed specified; I put it in the docs comment on the public API `validate_custom_proposal` . I also renamed the feature flag to `"prior_epoch_membership_key"` to make it a bit more clear. And force-pushed to rebase onto main head. 
Let me know if you'd like me to update this anywhere else!

I think your comment about the MLS encoding makes sense too. Would it make sense to do that change in a different PR? It could be used in other places in the mls-rs codebase that also include feature-flagged out fields in structs that derive `MlsEncode`.

mls-rs/src/group/epoch.rs
tomleavy: Good point, we do have to clean that up overall, can target for v 1.0

mulmarta (approved):

PR #265, opened 2025-04-09
Implementing `SelfRemove` proposal MLS extension

emarteca:
This PR adds an implementation of the `SelfRemove` proposal: this is a draft MLS extension, and is described [in this linked documentation](https://www.ietf.org/archive/id/draft-ietf-mls-extensions-06.html#name-selfremove-proposal).

The main features of the `SelfRemove` proposal are:
* it is specified by an empty struct, and the node to remove is determined by inspecting the sender
* it can only be included in a commit by reference
* it requires an UpdatePath
* it is allowed in External Commits

When a `SelfRemove` proposal is being included in an external commit being built by an external joiner, note that it validates the proposal before including it by reference but it cannot validate the membership tag (the MAC) because the external joiner is a non-member and does not have access to the membership. This limitation is specified in the MLS draft extension documentation.

During commit creation, if there is both a self-remove proposal and a remove proposal for the same node, the commit build fails with a `MlsError::MoreThanOneProposalForLeaf` error.

For more details, refer to the MLS draft extension docs linked above.

---
Note: in the [GSMA RCS specification](https://www.gsma.com/solutions-and-impact/technologies/networks/wp-content/uploads/2025/03/RCC.16-v1.0.pdf), section 7.11.8.1 specifies the self_remove proposal: this is equivalent to the MLS `SelfRemove` extension, however the extension value differs. In the MLS spec, the extension value is 8, and in the GSMA RCS spec the extension value is 0xF003. To accommodate both of these requirements, I have added 2 flags in this PR: `self_remove_proposal` and `rcs_self_remove_proposal`. If both flags are enabled, we use the GSMA RCS extension value, and with the `self_remove_proposal` flag only we get the MLS spec extension value.

### DIFF
--- a/mls-rs-core/Cargo.toml
+++ b/mls-rs-core/Cargo.toml
@@ -23,7 +23,6 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
 self_remove_proposal = []
-rcs_self_remove_proposal = []
 
 [dependencies]
 mls-rs-codec = { version = "0.6", path = "../mls-rs-codec", default-features = false}
--- a/mls-rs-core/Cargo.toml
+++ b/mls-rs-core/Cargo.toml
@@ -22,6 +22,8 @@ test_suite = ["serde", "dep:serde_json", "dep:itertools"]
 serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
+self_remove_proposal = []
+rfc_self_remove_proposal = []
 
 [dependencies]
 mls-rs-codec = { version = "0.6", path = "../mls-rs-codec", default-features = false}
--- a/mls-rs-core/Cargo.toml
+++ b/mls-rs-core/Cargo.toml
@@ -23,7 +23,7 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
 self_remove_proposal = []
-rfc_self_remove_proposal = []
+rcs_self_remove_proposal = []
 
 [dependencies]
 mls-rs-codec = { version = "0.6", path = "../mls-rs-codec", default-features = false}
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -56,7 +56,7 @@ impl ProposalType {
     pub const RE_INIT: ProposalType = ProposalType(5);
     pub const EXTERNAL_INIT: ProposalType = ProposalType(6);
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
-    pub const SELF_REMOVE: ProposalType = ProposalType(8);
+    pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
 
     /// Default proposal types defined
     /// in [RFC 9420](https://www.rfc-editor.org/rfc/rfc9420.html#name-leaf-node-contents)
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -56,6 +56,7 @@ impl ProposalType {
     pub const RE_INIT: ProposalType = ProposalType(5);
     pub const EXTERNAL_INIT: ProposalType = ProposalType(6);
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
+    pub const SELF_REMOVE: ProposalType = ProposalType(8);
 
     /// Default proposal types defined
     /// in [RFC 9420](https://www.rfc-editor.org/rfc/rfc9420.html#name-leaf-node-contents)
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -56,10 +56,10 @@ impl ProposalType {
     pub const RE_INIT: ProposalType = ProposalType(5);
     pub const EXTERNAL_INIT: ProposalType = ProposalType(6);
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
-    #[cfg(all(feature = "rfc_self_remove_proposal", feature = "self_remove_proposal"))]
+    #[cfg(all(feature = "rcs_self_remove_proposal", feature = "self_remove_proposal"))]
     pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
     #[cfg(all(
-        not(feature = "rfc_self_remove_proposal"),
+        not(feature = "rcs_self_remove_proposal"),
         feature = "self_remove_proposal"
     ))]
     pub const SELF_REMOVE: ProposalType = ProposalType(8);
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -56,13 +56,8 @@ impl ProposalType {
     pub const RE_INIT: ProposalType = ProposalType(5);
     pub const EXTERNAL_INIT: ProposalType = ProposalType(6);
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
-    #[cfg(all(feature = "rcs_self_remove_proposal", feature = "self_remove_proposal"))]
+    #[cfg(feature = "self_remove_proposal")]
     pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
-    #[cfg(all(
-        not(feature = "rcs_self_remove_proposal"),
-        feature = "self_remove_proposal"
-    ))]
-    pub const SELF_REMOVE: ProposalType = ProposalType(8);
 
     /// Default proposal types defined
     /// in [RFC 9420](https://www.rfc-editor.org/rfc/rfc9420.html#name-leaf-node-contents)
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -56,7 +56,13 @@ impl ProposalType {
     pub const RE_INIT: ProposalType = ProposalType(5);
     pub const EXTERNAL_INIT: ProposalType = ProposalType(6);
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
+    #[cfg(all(feature = "rfc_self_remove_proposal", feature = "self_remove_proposal"))]
     pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
+    #[cfg(all(
+        not(feature = "rfc_self_remove_proposal"),
+        feature = "self_remove_proposal"
+    ))]
+    pub const SELF_REMOVE: ProposalType = ProposalType(8);
 
     /// Default proposal types defined
     /// in [RFC 9420](https://www.rfc-editor.org/rfc/rfc9420.html#name-leaf-node-contents)
--- a/mls-rs/src/group/message_processor.rs
+++ b/mls-rs/src/group/message_processor.rs
@@ -2,6 +2,12 @@
 // Copyright by contributors to this project.
 // SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
+#[cfg(all(
+    feature = "by_ref_proposal",
+    feature = "custom_proposal",
+    feature = "self_remove_proposal"
+))]
+use super::SelfRemoveProposal;
 use super::{
     commit_sender,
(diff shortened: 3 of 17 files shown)


mls-rs-core/src/group/proposal_type.rs
```diff
@@ -56,6 +56,13 @@ impl ProposalType {
     pub const RE_INIT: ProposalType = ProposalType(5);
     pub const EXTERNAL_INIT: ProposalType = ProposalType(6);
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
+    #[cfg(all(feature = "rcs_self_remove_proposal", feature = "self_remove_proposal"))]
+    pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
+    #[cfg(all(
+        not(feature = "rcs_self_remove_proposal"),
+        feature = "self_remove_proposal"
+    ))]
+    pub const SELF_REMOVE: ProposalType = ProposalType(8);
```
mgeisler, line 65: Just a note: I don't think using the same name for two different constants is a good idea. I get that they model the same (or very very similar?) ideas, but constants are meant to be constants :smile:

Imagine if one of them drifts away from the other later — now what are people using mls-rs supposed to do?

emarteca: Hey -- you'll note these constants are mutually exclusive. This is just the ProposalType value for the `SelfRemove` proposal type. You'll see in the code that `rcs_self_remove_proposal` flag is only used in this one place. If this flag and `self_remove_proposal` are both enabled, then we get the proposal type value specified in the RCS GSMA spec for `SelfRemove`. If only the `self_remove_proposal` flag is used, then we get the proposal type value in the MLS draft spec for `SelfRemove`. Everything else with the `SelfRemove` implementation is behind the `self_remove_proposal` flag so it's the same whether or not we enable the `rcs_self_remove_proposal` flag.

I'm not a huge fan of this either, but it does not seem like a risk to me. These constants can never exist concurrently, it's one or the other. No functionality changes with the value. This is the only way I can think of to be able to satisfy both the RCS GSMA spec and the MLS draft spec which have different values for the `SelfRemove` proposal type.

mgeisler: > This is the only way I can think of to be able to satisfy both the RCS GSMA spec and the MLS draft spec which have different values for the `SelfRemove` proposal type.

I think this is a different way of saying that the two extensions are _different extensions_.

Imagine an application that wants to process both kinds of proposals. Or consider a library which uses mls-rs internally and wants to be flexible enough to process both extensions (by doing the same thing in each case, but it still needs to handle both values). Having a single constant with two values means such applications and libraries cannot easily exist (unless I'm missing something).

mulmarta: I actually don't think we should be using 8... It is only "suggested" in the IETF draft, and that same draft suggests 8 for a different proposal called AppDataUpdate. F003 is from the range "reserved for private use" which sounds proper until a type for SelfRemove is officially registered in IANA?

mls-rs-core/Cargo.toml
```diff
@@ -22,6 +22,8 @@ test_suite = ["serde", "dep:serde_json", "dep:itertools"]
 serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
+self_remove_proposal = []
+rcs_self_remove_proposal = []
```
tomleavy, line 26: I think it makes sense that we just have an rcs feature that can incorporate anything else that is needed moving forward as well?

mls-rs-core/src/group/proposal_type.rs
emarteca: Ah, well, if you are cool with just using `0xF003` that is very clean! Thanks.

mls-rs-core/Cargo.toml
emarteca: That makes sense! Based on @mulmarta 's comment above, if we use `0xF003` for the self-remove proposal type in all cases, then we don't need this flag in my PR anymore. Would you prefer that I still add it, and rename it to something more general like `rcs_spec_features` ?

tomleavy: Ok, if you don't need it anymore remove it and just note that if we do need this later on it should be general to reduce how many of these flags there are. 

Thanks!

mls-rs/src/group/mod.rs
```diff
@@ -4729,6 +4789,419 @@ mod tests {
...
+            .with_random_signing_identity("alice", TEST_CIPHER_SUITE)
+            .await
+            .custom_proposal_type(ProposalType::SELF_REMOVE)
+            .build()
+            .create_group(ExtensionList::new(), Default::default())
+            .await
+            .unwrap();
+
+        let (bob_signing_identity, bob_secret_key) =
+            get_test_signing_identity(TEST_CIPHER_SUITE, b"bob").await;
```
mulmarta, line 4831: [nit] Can `self_remove_group_setup` be used?

mls-rs/src/group/mod.rs
```diff
@@ -4729,6 +4789,419 @@ mod tests {
...
+            .process_incoming_message(bob_self_remove.clone())
+            .await
+            .unwrap();
+
+        // Alice also removes Bob with a regular remove proposal.
+        alice.propose_remove(1, Vec::new()).await.unwrap();
+
+        // Alice commits Bob's self-remove and Alice's removal of Bob. This is an error.
+        let commit = alice.commit(Vec::new()).await;
+        assert_matches!(commit, Err(MlsError::MoreThanOneProposalForLeaf(1)));
```
mulmarta, line 4945: To be consistent with handling of other proposals, the commit should succeed but one of the conflicting proposals should be filtered out. This is so that you don't get stuck - right now Alice can't ever commit unless she clears proposals.

mls-rs/src/group/proposal.rs
```diff
@@ -232,6 +232,29 @@ impl Debug for ExternalInit {
...
+#[cfg(all(
+    feature = "by_ref_proposal",
+    feature = "custom_proposal",
+    feature = "self_remove_proposal"
+))]
+impl Debug for SelfRemoveProposal {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SelfRemoveProposal").finish()
+    }
+}
```
mulmarta, line 256: [nit] You can just derive Debug, we implement manually above only to use `pretty_bytes`

mls-rs/src/group/proposal_filter/filtering.rs
```diff
@@ -106,6 +113,13 @@ where
         let proposals = filter_out_reinit_if_other_proposals(strategy.is_ignore(), proposals)?;
         let proposals = filter_out_external_init(strategy, proposals)?;
 
+        #[cfg(all(
+            feature = "by_ref_proposal",
+            feature = "custom_proposal",
+            feature = "self_remove_proposal"
+        ))]
+        check_for_remove_and_self_remove_same_leaf(&proposals)?;
```
mulmarta, line 121: This is where it should filter out proposal or error out according to `strategy`. See `filter_out_external_init` above. BTW a commit with remove and self remove would fail on `batch_edit` anyway - `blank_leaf_node` function I think fails if it's called on a leaf twice.

mls-rs/src/group/proposal_filter/filtering.rs
```diff
@@ -460,34 +517,58 @@ pub(crate) fn proposer_can_propose(
...
+                    | ProposalType::RE_INIT
+                    | ProposalType::GROUP_CONTEXT_EXTENSIONS
+            );
+            #[cfg(all(
+                feature = "by_ref_proposal",
+                feature = "custom_proposal",
+                feature = "self_remove_proposal"
+            ))]
+            let can_propose = can_propose || matches!(proposal_type, ProposalType::SELF_REMOVE);
+            can_propose
```
mulmarta, line 535: I don't think self remove can be by value - that would cause "committer can't self remove"? But it could be local I guess

mls-rs/src/group/proposal_filter/filtering.rs
```diff
@@ -460,34 +517,58 @@ pub(crate) fn proposer_can_propose(
...
+                proposal_type,
+                ProposalType::ADD
+                    | ProposalType::REMOVE
+                    | ProposalType::RE_INIT
+                    | ProposalType::PSK
+                    | ProposalType::GROUP_CONTEXT_EXTENSIONS
+            );
+
+            #[cfg(all(feature = "custom_proposal", feature = "self_remove_proposal"))]
+            let can_propose = can_propose || matches!(proposal_type, ProposalType::SELF_REMOVE);
```
mulmarta, line 569: I don't think external committer can propose  (its own) self-remove for the same reason as above?

mls-rs/src/tree_kem/mod.rs
```diff
@@ -344,6 +351,35 @@ impl TreeKemPublic {
...
+                    identity(&old_leaf.signing_identity, id_provider, extensions).await?;
+
+                self.index.remove(old_leaf, &identity);
+            }
+
+            if proposal_bundle.self_removes[i].is_by_value() || !filter {
+                res?;
+            } else if res.is_err() {
+                proposal_bundle.remove::<SelfRemoveProposal>(i);
+            }
```
mulmarta, line 381: This code is almost the same as below - can the duplication be avoided somehow (function / chain removes and self-removes)?

mulmarta (commented):
Thanks a lot for adding the feature!

mls-rs-core/src/group/proposal_type.rs
emarteca: Updated; resolving.

mls-rs-core/Cargo.toml
emarteca: Makes sense!

mls-rs/src/group/proposal_filter/filtering.rs
emarteca: Ah that's a good point; I've removed it. Thanks.

mls-rs/src/group/proposal_filter/filtering.rs
emarteca: Ah good point! Youre totally right. I've split up the `Local` and `ByValue` cases, and removed the self-remove proposal from the `ByValue` case

mls-rs/src/group/proposal.rs
emarteca: Ah, makes sense! Thanks! Done.

mls-rs/src/group/message_processor.rs
```diff
@@ -707,7 +720,30 @@ pub(crate) trait MessageProcessor: Send + Sync {
...
+                *removed.proposal.to_remove,
+            ));
+        }
         let is_self_removed = self_removed.is_some();
+        #[cfg(all(
+            feature = "by_ref_proposal",
+            feature = "custom_proposal",
+            feature = "self_remove_proposal"
+        ))]
+        let is_self_removed = is_self_removed || self_removed_by_self.is_some();
```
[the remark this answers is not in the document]
emarteca: Ah, great catch! Thanks.

mls-rs/src/group/message_processor.rs
```diff
@@ -707,7 +720,30 @@ pub(crate) trait MessageProcessor: Send + Sync {
...
+        #[cfg(all(
+            feature = "by_ref_proposal",
+            feature = "custom_proposal",
+            feature = "self_remove_proposal"
+        ))]
+        if let (Some(removed), Some(_)) = (&self_removed, &self_removed_by_self) {
+            return Err(MlsError::MoreThanOneProposalForLeaf(
+                *removed.proposal.to_remove,
+            ));
+        }
```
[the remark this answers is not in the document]
emarteca: Ah you're right, it does (through `apply_proposals` --> `apply_proposals_from_member`. Good point.

mls-rs/src/group/proposal_filter/filtering.rs
emarteca: Ah that makes sense. Done!

mls-rs/src/group/mod.rs
emarteca: Good point -- done with your suggested filter change in the proposal filtering

mls-rs/src/group/mod.rs
emarteca: Yes; done!

mls-rs/src/tree_kem/mod.rs
emarteca: Good point! Done.

emarteca:
Thank you for all the comments! I've applied edits; force-pushed to rebase onto main HEAD.

tomleavy (approved):
Thanks!

mulmarta (approved):
Thank you!

emarteca:
Thank you!! :pray:

PR #273, opened 2025-04-25
Adding a trait to provide current time through the client config

emarteca:
There are various places in the source code currently where `MlsTime::now()` is called. This calls the `SystemTime::now` (unless we are running in WASM).

This PR adds the ability to specialize the clock. This is useful both for testing (setting a deterministic clock), and also for allowing an application to provide an external clock.

### DIFF
--- a/mls-rs-core/src/time.rs
+++ b/mls-rs-core/src/time.rs
@@ -60,7 +60,6 @@ impl MlsTime {
 #[derive(Clone, Debug, Default)]
 pub struct DefaultCurrentTime {}
 
-#[cfg(any(target_arch = "wasm32", feature = "std"))]
 impl CurrentTimeProvider for DefaultCurrentTime {
     fn get_current_time_seconds(&self) -> u64 {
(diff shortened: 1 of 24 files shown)


mulmarta:
Hey @emarteca ! I'm a bit hesitant to add yet more trait - something we explicitly want to avoid in the 1.x branch... Going through the code, `now()` is not called in too many places. Maybe using the builder pattern plus explicit inputs we can make the API clean?
- tests -- calling this in tests seems wrong, it should be some constant?
- commit -- we can have optional custom time in `CommitBuilder`
- validate key package -- we can have input
- key package lifetime -- custom time can be input for now, I think 1.x already has KeyPackageBuilder

emarteca:
Thanks for the review -- and please excuse the delay in response! That makes sense. I will refactor and ping back with changes soon.

emarteca:
Thanks again for your review -- I've refactored the code as you suggested, in https://github.com/awslabs/mls-rs/pull/282

PR #274, opened 2025-04-26
Fixing missing update_hash for SelfRemove proposals

emarteca:
Bug fix; missing updating the hashes on self-removed leaves in `batch_edit` . This caused a `TreeHashMismatch` error on construction of an external commit with the GroupInfo after processing a commit with a self-remove proposal, because the tree hash in the group context did not match the true tree hash.

This PR fixes the bug and adds a test that was repro'ing the previous bug.


mulmarta (approved):

stefunctional (approved):
Thank you!

PR #275, opened 2025-04-26
Allow `process_incoming_message_with_time` to work on a commit from self

emarteca:
Without special handling for a pending commit, `process_incoming_message_with_time` would fail on a commit from self. This PR just copies the logic for handling pending messages from self, from `process_incoming_message`, into `process_incoming_message_with_time`, and adds a test. Without the change, the added test would fail with `CantProcessMessageFromSelf` [here](https://github.com/awslabs/mls-rs/blob/main/mls-rs/src/group/mod.rs#L2125).


stefunctional (approved):
Thank you!

PR #280, opened 2025-05-01
Server remove proposal

emarteca:
Implementation of the server_remove proposal, corresponding to section 7.11.9 of the [GSMA RCS spec](https://www.gsma.com/solutions-and-impact/technologies/networks/wp-content/uploads/2025/03/RCC.16-v1.0.pdf). 

This proposal represents removal prompted by the server. A group member must still both propose and commit it (since the server is not a group member), but it signals that the removal was requested by the server. 

The server_remove is a proposal that has the following properties: 
* Extension Value: 0xF004
* Extension Name: server_remove
* External: Yes
* Path Required: Yes

It effectively has all of the same properties and behaves identically to the existing `Remove` proposal, except that multiple of them can be committed externally.

### DIFF
--- a/mls-rs-core/Cargo.toml
+++ b/mls-rs-core/Cargo.toml
@@ -23,6 +23,7 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
 self_remove_proposal = []
+server_remove_proposal = []
 
 [dependencies]
 mls-rs-codec = { version = "0.6", path = "../mls-rs-codec", default-features = false}
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -1078,6 +1078,24 @@ where
         self.proposal_message(proposal, authenticated_data).await
     }
 
+    #[cfg_attr(feature = "ffi", safer_ffi_gen::safer_ffi_gen_ignore)]
+    #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    pub async fn propose_server_remove(
+        &mut self,
+        index: u32,
+        authenticated_data: Vec<u8>,
+    ) -> Result<MlsMessage, MlsError> {
+        let leaf_index = LeafIndex(index);
+
+        // Verify that this leaf is actually in the tree
+        self.current_epoch_tree().get_leaf_node(leaf_index)?;
+
+        let proposal = Proposal::ServerRemove(ServerRemoveProposal {
+            to_remove: leaf_index,
+        });
+        self.proposal_message(proposal, authenticated_data).await
+    }
+
     /// Create a proposal message that adds an external pre shared key to the group.
     ///
     /// Each group member will need to have the PSK associated with
@@ -2290,6 +2308,18 @@ where
             .self_removes
             .iter()
             .find(|p| p.sender == Sender::Member(*self.private_tree.self_index))
+            .cloned()
+    }
+
+    fn server_removal_proposal(
+        &self,
+        provisional_state: &ProvisionalState,
+    ) -> Option<ProposalInfo<ServerRemoveProposal>> {
+        provisional_state
+            .applied_proposals
+            .server_removes
+            .iter()
+            .find(|p| p.proposal.to_remove == self.private_tree.self_index)
             .cloned()
     }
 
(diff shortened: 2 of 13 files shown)


mls-rs-core/Cargo.toml
```diff
@@ -23,6 +23,7 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
 self_remove_proposal = []
+server_remove_proposal = []
```
mulmarta, line 26: Should there be just a GSMA feature?

mls-rs/src/group/proposal_filter/filtering.rs
```diff
@@ -342,6 +347,39 @@ fn filter_out_remove_if_self_remove_same_leaf(
...
+        )
+    })?;
+    Ok(proposals)
+}
+
+#[cfg(feature = "server_remove_proposal")]
+fn filter_out_server_remove_if_remove_same_leaf(
+    strategy: FilterStrategy,
+    mut proposals: ProposalBundle,
+) -> Result<ProposalBundle, MlsError> {
```
mulmarta, line 368: That shouldn't happen, no? Server cannot have a leaf or be a committer?

mulmarta (commented):
I have two general questions:
1. How is server remove different than normal MLS remove sent by an external sender?
2. Is serialization the reason why the mls-rs custom proposal mechanism cannot be used for this?

mls-rs-core/Cargo.toml
emarteca: Yes, that is a good idea.

mls-rs/src/group/mod.rs
```diff
@@ -1078,6 +1078,31 @@ where
...
+    #[cfg_attr(feature = "ffi", safer_ffi_gen::safer_ffi_gen_ignore)]
+    fn server_remove_proposal(&self, index: u32) -> Result<Proposal, MlsError> {
+        let leaf_index = LeafIndex(index);
+
+        // Verify that this leaf is actually in the tree
+        self.current_epoch_tree().get_leaf_node(leaf_index)?;
+
+        Ok(Proposal::ServerRemove(ServerRemoveProposal {
+            to_remove: leaf_index,
+        }))
```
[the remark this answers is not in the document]
emarteca: So the logic flow is a bit odd; but in this model the server is not generating the proposal. Rather, the server signals to the client that the client should remove a member, and then the client has to create the proposal. It's more of a "server said to remove" remove proposal

emarteca:
Thanks for the comments, and (like on my other PR), so sorry for the delay in response! 

To answer your questions:
1. In this case, server remove is not actually issued by the server. Rather, the server signals to the client that it should remove a member, but flag it as having been done by the server. It's basically the same code as the Remove proposal, except that there can be more than one in an external commit.
2. My understanding of the code is that custom proposals could not add/remove members from the tree directly, and we need this proposal to remove the member when it is processed. (basically, we need it to be the exact same as the Remove proposal except with a different ProposalType, and where more than one are allowed in an external commit). Maybe I'm missing something here though?

I'll followup with code changes soon, thanks again!

mls-rs-core/Cargo.toml
emarteca: done; resolving

emarteca:
> Custom proposals definitely can add/remove members from the tree. Like in the [example]
Oh that's great! I didn't know about this, thank you! I will try this out and likely abandon this PR :smile: 

> Why does the GSMA sec say "External: Yes"? According to the [RFC](https://www.rfc-editor.org/rfc/rfc9420.html#section-17.4-3.4) this means that servers can issue this proposal. It does not have anything to do with external committers - e.g. external_init has "External: No" and it's mandatory for an external commit.
Good points; we are considering this and are in internal discussion about this with the relevant parties.

> What prevents a malicious member Alice from sending a commit with server_remove of Bob that the server never requested?
Nothing. However, in an RCS MLS 1.0 ecosystem we lack the root of trust infrastructure to be able to obtain the authorized external senders. Because of this, the server operations are, in MLS 1.0, unauthenticated. So we don't want the server to be able to generate proposals, and do not support external senders.

mulmarta:
@emarteca what's the status on this one? Did you manage to implement it with custom proposal, or is there an issue?

PR #282, opened 2025-05-20
Adding custom time through commit builder and various explicit inputs

emarteca:
Refactoring https://github.com/awslabs/mls-rs/pull/273 to use `CommitBuilder` and explicit inputs instead of the time trait.  Making another branch and separate PR because it's cleaner than redoing the previous branch; will close both PRs when one is resolved.

Thanks for the recommended changes!

### DIFF
--- a/mls-rs/src/client.rs
+++ b/mls-rs/src/client.rs
@@ -20,6 +20,7 @@ use crate::group::{
 use crate::identity::SigningIdentity;
 use crate::key_package::{KeyPackageGeneration, KeyPackageGenerator};
 use crate::protocol_version::ProtocolVersion;
+use crate::time::MlsTime;
 use crate::tree_kem::node::NodeIndex;
 use alloc::vec::Vec;
 use mls_rs_codec::MlsDecode;
@@ -410,12 +411,13 @@ where
     }
 
     #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
-    pub fn to_builder(&self) -> ClientBuilder<MakeConfig<C>> {
+    pub fn to_builder(&self, maybe_now_time: Option<MlsTime>) -> ClientBuilder<MakeConfig<C>> {
         ClientBuilder::from_config(recreate_config(
             self.config.clone(),
             self.signer.clone(),
             self.signing_identity.clone(),
             self.version,
+            maybe_now_time,
         ))
     }
 
@@ -433,13 +435,15 @@ where
     ///
     /// A key package message may only be used once.
     #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
     pub async fn generate_key_package_message(
         &self,
         key_package_extensions: ExtensionList,
         leaf_node_extensions: ExtensionList,
+        maybe_now_time: Option<MlsTime>,
     ) -> Result<MlsMessage, MlsError> {
         Ok(self
-            .generate_key_package(key_package_extensions, leaf_node_extensions)
+            .generate_key_package(key_package_extensions, leaf_node_extensions, maybe_now_time)
             .await?
             .key_package_message())
     }
@@ -449,6 +453,7 @@ where
         &self,
         key_package_extensions: ExtensionList,
         leaf_node_extensions: ExtensionList,
+        maybe_now_time: Option<MlsTime>,
     ) -> Result<KeyPackageGeneration, MlsError> {
         let (signing_identity, cipher_suite) = self.signing_identity()?;
 
@@ -467,7 +472,7 @@ where
 
         let key_pkg_gen = key_package_generator
             .generate(
-                self.config.lifetime(),
+                self.config.lifetime(maybe_now_time),
                 self.config.capabilities(),
                 key_package_extensions,
                 leaf_node_extensions,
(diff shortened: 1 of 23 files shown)


mls-rs/src/group/mod.rs
```diff
@@ -308,7 +309,7 @@ where
             config.leaf_properties(leaf_node_extensions),
             signing_identity,
             &signer,
-            config.lifetime(),
+            config.lifetime(maybe_now_time),
```
mgeisler, line 312: Can you also pass the timestamp to `ValidationContext::Add(None)` below? That is, change

```rust
        leaf_node_validator
            .check_if_valid(
                public_tree.get_leaf_node(LeafIndex(0))?,
                ValidationContext::Add(None),
            )
            .await?;
```

to

```rust
        leaf_node_validator
            .check_if_valid(
                public_tree.get_leaf_node(LeafIndex(0))?,
                ValidationContext::Add(maybe_now_time),
            )
            .await?;
```

That will mean we validate that the group creator has a valid credential at the time of group creation.

PS, I would not prefix the variable with `maybe_`, I would just call it `timestamp` to be consistent with the naming in [`IdentityProvider`](https://docs.rs/mls-rs/latest/mls_rs/trait.IdentityProvider.html).

emarteca: yes, makes sense! done.

mls-rs/src/client.rs
```diff
@@ -410,12 +411,13 @@ where
     }
 
     #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
-    pub fn to_builder(&self) -> ClientBuilder<MakeConfig<C>> {
+    pub fn to_builder(&self, maybe_now_time: Option<MlsTime>) -> ClientBuilder<MakeConfig<C>> {
```
mgeisler, line 414: I would suggest aligning with the existing code and call this `timestamp` consistently throughout:

```suggestion
    pub fn to_builder(&self, timestamp: Option<MlsTime>) -> ClientBuilder<MakeConfig<C>> {
```

mls-rs/src/client.rs
```diff
@@ -433,13 +435,15 @@ where
     ///
     /// A key package message may only be used once.
     #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    #[cfg_attr(all(feature = "ffi", not(test)), safer_ffi_gen::safer_ffi_gen_ignore)]
     pub async fn generate_key_package_message(
         &self,
         key_package_extensions: ExtensionList,
         leaf_node_extensions: ExtensionList,
+        maybe_now_time: Option<MlsTime>,
```
mgeisler, line 443: Looking at the code, I find the interaction between an `Option<MlsTime>` and the actual key package generation confusing: the code really really wants an actual timestamp to use as the starting point for the lifetime.

So this braking API change would a chance to give it that: require people to pass in a `MlsTime` here and then the calls to `MlsTime::now()` below can go away! The conditional compilation code for Wasm will go away too.

mls-rs/src/group/commit.rs
```diff
@@ -523,6 +537,12 @@ where
         #[cfg(not(feature = "std"))]
         let time = None;
 
+        let time = if commit_time.is_some() {
```
mgeisler, line 540: Here is another place that can be simplified by requiring a `MlsTime` instead of `Option<MlsTime>` above.

mls-rs/src/tree_kem/lifetime.rs
```diff
@@ -22,13 +22,19 @@ impl Lifetime {
...
         #[cfg(not(feature = "std"))]
         // There is no clock on no_std, this is here just so that we can run tests.
         let not_before = 3600u64;
 
+        let not_before = if let Some(not_before_time) = maybe_not_before {
+            not_before_time.seconds_since_epoch()
+        } else {
+            not_before
+        };
+
```
mgeisler, line 37: Another case that would benefit from not having an optional `MlsTime`.

mgeisler:
Hi @tomleavy, thanks for looking at this!

> get this to a place you guys are happy with and then I can review

We've been using this internally for a few weeks now and I believe it has fixed our previous problems with tests that rely on "now" being, well, now :smile:

mls-rs/src/client.rs
emarteca: Makes sense; done.

emarteca:
Force-pushed to rebase onto main HEAD; fixed merge conflicts. Thanks for the review!

Regarding @mgeisler 's comments about requiring time instead of making it optional -- good point, I agree that would be cleaner. I played around with this change for a bit, but in trying to propagate a required MlsTime in `generate_key_package` I found that this leads to a much bigger refactor. E.g., it would then _require_ anyone building a commit to provide the commit_time, since key packages are generated internally through some code paths. 

If I'm going to do this bigger refactor, and remove all calls to `MlsTime::now()` in the library, then I think we'll need to think a bit more about this (e.g., potentially remove support for `process_incoming_message` and only allow `process_incoming_message_with_time` ?), and sketch out all of the places where time could still be optional. I'd rather do this in a followup PR.

Echoing @mgeisler , internally we've been using this PR and it provides us the time-validation support we need.

Thanks again for your time and review on this!

mulmarta (approved):

mgeisler:
> E.g., it would then _require_ anyone building a commit to provide the commit_time, since key packages are generated internally through some code paths.

Could those code paths not just use `None`? Or are you saying that the code paths use "now" today? If so, it sounds like they need to be updated as well (unfortunately).

tomleavy (approved):

PR #284, opened 2025-05-30
Use `MlsTime` more consistently

mgeisler:
### Description of changes:

This changes the `Lifetime` struct to use `MlsTime` instead of `u64` to represent the fixed `not_before` and `not_after` timestamps.

This change came out of an internal discussion about what the unit is for the field: does it measure a different from “now”, or is it an absolute place in time? Turns out it’s the latter and this is now encoded in the type system.

### Testing:

Existing tests were updated.


mgeisler:
Thanks Tom! @mulmarta, would you be up for reviewing this one too?

stefunctional (approved):

PR #285, opened 2025-06-06
Authenticate key generation in application messages.

cindylindeed:
Defends against in-group forgery attacks where group members can replay messages sent by other members, see https://eprint.iacr.org/2025/554.

### Description of changes:

* Adds `GenerationAuthMode` to `EncryptionOptions`. `GenerationAuthMode` defaults to `None`, so no default clients are modified.
* Implements adding key generation in plaintext as the first 4 bytes of the authenticated data, which is signed by the sender and thereby authenticated.
* Verify the key generation returned from the ratchet matches the one in the authenticated data in AEAD seal / open.

### Call-outs:

* mls-rs/src/group/generation_authentication.rs
* mls-rs/src/group/mod.rs
* mls-rs/src/group/ciphertext_processor.rs

### Testing:

Unit tests.


mls-rs/src/group/ciphertext_processor.rs
```diff
@@ -146,6 +148,12 @@ where
         let key_data = self.next_encryption_key(key_type).await?;
         let generation = key_data.generation;
 
+        // Depending on the GenerationAuthMode, verify the key generation obtained from
+        // the ratchet matches the key generation prefix in `authenticated_data`
+        if content_type == ContentType::Application {
```
mgeisler, line 153: I see this check is there at least twice. Would it make sense to push this down to the `verify_generation` method?

That could simplify the callers and perhaps ensure that the check is always done — if that is important?

mls-rs/src/group/ciphertext_processor.rs
```diff
@@ -298,19 +315,24 @@ mod test {
     }
 
     #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
-    async fn test_data(cipher_suite: CipherSuite) -> TestData {
+    async fn test_data(cipher_suite: CipherSuite, generation_auth: GenerationAuthMode) -> TestData {
```
mgeisler, line 318: Here and elsewhere: why is the variable named `..._auth`, but the type is `...AuthMode`?

Since `GenerationAuthMode` is an enum, it's kind of implicit that it selects a "mode", no? Perhaps it could lose the suffix then. Just a thought, not very critical, but it caught my eye that you're not consistent in the naming :smile:

mls-rs/src/group/generation_authentication.rs
```diff
@@ -0,0 +1,84 @@
...
+        authenticated_data: &[u8],
+    ) -> Result<(), MlsError> {
+        match self {
+            GenerationAuthMode::None => Ok(()),
+
+            GenerationAuthMode::Plaintext => {
+                if authenticated_data.len() < 4 {
+                    return Err(MlsError::MissingAuthGeneration);
+                }
+                let auth_generation = u32::from_be_bytes(authenticated_data[0..4].try_into()?);
```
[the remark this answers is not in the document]
cindylindeed: This panics when the index is out of range: `range end index 4 out of range for slice of length 0`.

mls-rs/src/client.rs
```diff
@@ -35,6 +35,12 @@ use crate::group::external_commit::ExternalCommitBuilder;
 #[cfg(feature = "by_ref_proposal")]
 use alloc::boxed::Box;
 
+#[cfg(feature = "std")]
+use std::array::TryFromSliceError;
+
+#[cfg(not(feature = "std"))]
+use core::array::TryFromSliceError;
```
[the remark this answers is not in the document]
cindylindeed: Deleted.

mls-rs/src/client.rs
```diff
@@ -365,6 +383,13 @@ impl From<ExtensionError> for MlsError {
     }
 }
 
+impl From<TryFromSliceError> for MlsError {
```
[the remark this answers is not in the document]
cindylindeed: Done.

mls-rs/src/group/ciphertext_processor.rs
```diff
@@ -146,6 +148,12 @@ where
         let key_data = self.next_encryption_key(key_type).await?;
         let generation = key_data.generation;
 
+        // Depending on the GenerationAuthMode, verify the key generation obtained from
+        // the ratchet matches the key generation prefix in `authenticated_data`
+        if content_type == ContentType::Application {
+            generation_auth.verify_generation(generation, &authenticated_data)?;
```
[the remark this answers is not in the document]
cindylindeed: Done.

mls-rs/src/group/ciphertext_processor.rs
cindylindeed: I want to add generation authentication to private handshake messages in an upcoming PR, so this logic makes sense here vs buried in `GenerationAuthMode`. In the subsequent PR, this if statement would be removed.

mls-rs/src/group/ciphertext_processor.rs
cindylindeed: Done, originally named `GenerationAuthMode` to match the sibling `PaddingMode` defined in `EncryptionOptions`.

mls-rs/src/group/generation_authentication.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl GenerationAuthMode {
+    /// Prepend the key generation to `authenticated_data` based on the mode.
+    pub(super) fn prepend_generation<GS: GroupStateProvider, CS: CipherSuiteProvider>(
+        &self,
+        group_state: &mut GS,
+        cipher_suite_provider: &CS,
+        content_type: ContentType,
+        authenticated_data: &mut Vec<u8>,
+    ) -> Result<Vec<u8>, MlsError> {
```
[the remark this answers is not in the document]
cindylindeed: Changed the input parameters and the way the vectors are concatenated; PTAL.

mls-rs/src/group/generation_authentication.rs
```diff
@@ -0,0 +1,84 @@
...
+        match self {
+            GenerationAuthMode::None => Ok(authenticated_data.to_vec()),
```
mgeisler, line 39: I believe this is already a `Vev`, so this is probably just `clone` call in disguise? The Clippy linter should be able to tell.

mls-rs/src/group/generation_authentication.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl GenerationAuthMode {
+    /// Prepend the key generation to `authenticated_data` based on the mode.
+    pub(super) fn prepend_generation<GS: GroupStateProvider, CS: CipherSuiteProvider>(
+        &self,
+        group_state: &mut GS,
+        cipher_suite_provider: &CS,
+        content_type: ContentType,
+        authenticated_data: &mut Vec<u8>,
```
mgeisler, line 36: I'm on my phone, but I don't immediately see where this needs to be mutable?

As a rule of thumb, it's unusual to take a mutable reference as input. So it calls attention to itself 🙂

mulmarta:
Thanks @cindylindeed for the proposal!

A couple of general questions:

How do clients advertise support for different auth modes? That is, how does group member A know that all members support Plaintext auth mode so that A can safely use it?

In general, it may be possible that a client supports both modes, Plaintext and None, e.g. to talk to not-yet-updated clients. When a message comes in, how does the client know which mode to use? Using the wrong one can be dangerous, since e.g. 4 bytes of regular metadata may disguise as a generation or vice versa.

cindylindeed:
> How do clients advertise support for different auth modes? That is, how does group member A know that all members support Plaintext auth mode so that A can safely use it?
> 
> In general, it may be possible that a client supports both modes, Plaintext and None, e.g. to talk to not-yet-updated clients. When a message comes in, how does the client know which mode to use? Using the wrong one can be dangerous, since e.g. 4 bytes of regular metadata may disguise as a generation or vice versa.

Completely agree with this. We're planning to add the mode as a GroupContext Extension to achieve group agreement.

mls-rs/src/group/generation_authentication.rs
cindylindeed: Changed in newest commit that I hadn't pushed earlier.

mls-rs/src/group/generation_authentication.rs
cindylindeed: Changed in newest commit that I hadn't pushed earlier.

cindylindeed:
> Additional question about the other `WARNING` - what prevents us from encrypting the generation instead of including it in associated data?

Fair point, as the plaintext is also signed. There is a higher philosophical barrier (in my opinion) around mutating the plaintext vs the associated data (where the key generation fits more naturally by definition). But padding is also mutating the plaintext, so perhaps this sentiment doesn't hold. As long as the encoding is injective (which is required for adding to the associated data too), there isn't any issue with adding it to the plaintext.

Let me update the PR to reflect this. I had a think and do believe the pros outweigh the softer philosophical concerns. Thank you for the suggestion.

> Could this be part of this PR? So that when we receive a message, we use the auth mode from GCE instead of using a fixed property of the client? I feel like making it a fixed client property is not what we want in the end.

Yes, I'll work on this and push a commit. Replying first to get the discussion rolling.

Cheers~

mulmarta:
Thank you @cindylindeed ! (Good point about analogy to padding.)

PR #289, opened 2025-06-18
Include timestamps in `MlsError::InvalidLifetime`

mgeisler:
### Description of changes:

This makes it much easier to debug the error after the fact.

### Call-outs:

This is a breaking change, so I’m also bumping the mls-rs version number to reflect this. As always, we would appreciate a release soon after, if possible.

### Testing:

Existing tests were updated to assert on the new values (where possible: some values change with every test run since "now" isn't fixed).


tomleavy (approved):
Thanks!

mulmarta (approved):

PR #290, opened 2025-06-19
Report rows changed on SqLite inserts

JHabz:
### Description of changes:

Allows adding control behavior depending on if what is being inserted already exists at that key/value pair, or if some subset already exists for the case of transact_insert.

### Testing:

Modified and added tests to verify that for each of the inserts the number is reported correctly for new/unique and existing entries

### DIFF
--- a/mls-rs-ffi/Cargo.toml
+++ b/mls-rs-ffi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mls-rs-ffi"
-version = "0.10.0"
+version = "0.10.1"
 edition = "2021"
 description = "Helper crate to generate FFI definitions for mls-rs"
 homepage = "https://github.com/awslabs/mls-rs"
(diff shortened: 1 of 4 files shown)


mls-rs/Cargo.toml
```diff
@@ -67,7 +67,7 @@ spin = { version = "0.9.8", default-features = false, features = ["mutex", "spin
 maybe-async = { version = "0.2.10" }
 
 # Optional dependencies
-mls-rs-provider-sqlite = { path = "../mls-rs-provider-sqlite", version = "0.16.0", default-features = false, optional = true }
+mls-rs-provider-sqlite = { path = "../mls-rs-provider-sqlite", version = "0.17.0", default-features = false, optional = true }
```
stefunctional, line 70: This dependency is re-exported by `mls-rs`, which makes it part of the public API of `mls-rs`, which implies this is also a breaking change for `mls-rs` and its version needs to be bumped accordingly (`0.48.0`).

tomleavy (approved):

PR #291, opened 2025-06-23
Add docstring to `MlsTime`

mgeisler:
### Description of changes:

Since the difference between `MlsTime`, `SystemTime`, `Instant`, and `Duration` comes up regularly in internal code reviews, I figured it would be helpful to briefly document what `MlsTime` represents and why it can be trivially converted to/from a `Duration` even though the two types represent different things (point in time vs difference in time).

### Testing:

Only updates documentation.


stefunctional (approved):
Thank you!

PR #294, opened 2025-07-10
Export key generation during encryption, decryption of PrivateMessages.

cindylindeed:
### Description of changes:

Used by clients to authenticate the generation to defend against in-group forgery attacks described in https://eprint.iacr.org/2025/554. This may be accomplished by placing the generation in the `message` or `authenticated_data` parameters of `Group::encrypt_application_message`, as both fields are signed by the sender's signature key.

To verify, get the unauthenticated generation from ApplicationMessageDescription returned from `Group::process_incoming_message`, which is the value used to derive keys to decrypt the message, and check that it equals the authenticated generation.

Legacy PR https://github.com/awslabs/mls-rs/pull/285 appended and verified the authenticated and unauthenticated key generation values in `mls-rs` itself. That has been abandoned as we now prefer the client to choose where to put the generation.

### Call-outs:

UPDATE: The `mls-rs` owners prefer 1) the generation to be returned via ApplicationMessageDescription and 2) decryption of SenderData to occur twice inside the decrypt flow.

I have not started implementing key generation export on decrypt, as I wanted to check y'all's preference before starting that surgery. **Please let me know what you think re: decrypt options A and B.**

```
// encrypt

msg, ad := ...
gen := peek_next_key_generation()  // new
// put gen inside msg or ad
ciphertext := encrypt_application_message(msg, ad)

--

// decrypt option A

msg_and_ad, unauthn_gen := process_incoming_message_with_time_and_get_unathn_key_generation(ciphertext, time)  // new
// verify msg_and_ad.gen == unauthn_gen

// PRO: decrypts SenderData once
// CON: threading the key generation through the many internal API layers requires defining new protos (eg
// Group::decrypt_incoming_ciphertext returns AuthenticatedContent [1], which unauthn_key_generation
// cannot sit in, so AuthenticatedContentAndUnauthnKeyGeneration has to be created)

--

// decrypt option B

unauthn_gen := get_unauthn_key_generation(ciphertext)  // new
msg_and_ad := process_incoming_message_with_time(ciphertext, time)
// verify msg_and_ad.gen == unauthn_gen

// CON: decrypts SenderData twice
// PRO: more symmetric to encrypt flow
```

[1] https://github.com/awslabs/mls-rs/blob/d9e5f8bac5fd6a531bc8b403afdeef5e45aa589e/mls-rs/src/group/mod.rs#L1424

### Testing:

Unit tests.

### DIFF
--- a/mls-rs-core/Cargo.toml
+++ b/mls-rs-core/Cargo.toml
@@ -23,6 +23,7 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
 self_remove_proposal = []
+gsma_rcs_e2ee_feature = []
 
 [dependencies]
 mls-rs-codec = { version = "0.7", path = "../mls-rs-codec", default-features = false}
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -58,6 +58,10 @@ impl ProposalType {
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
     #[cfg(feature = "self_remove_proposal")]
     pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
+    #[cfg(feature = "gsma_rcs_e2ee_feature")]
+    pub const RCS_SIGNATURE: ProposalType = ProposalType::new(0xF002);
+    #[cfg(feature = "gsma_rcs_e2ee_feature")]
+    pub const RCS_SERVER_REMOVE: ProposalType = ProposalType::new(0xF004);
 
     /// Default proposal types defined
     /// in [RFC 9420](https://www.rfc-editor.org/rfc/rfc9420.html#name-leaf-node-contents)
--- a/mls-rs/Cargo.toml
+++ b/mls-rs/Cargo.toml
@@ -37,6 +37,7 @@ rfc_compliant = ["private_message", "custom_proposal", "out_of_order", "psk", "x
 last_resort_key_package_ext = ["mls-rs-core/last_resort_key_package_ext"]
 self_remove_proposal = ["mls-rs-core/self_remove_proposal"]
 export_key_generation = []
+gsma_rcs_e2ee_feature = ["mls-rs-core/gsma_rcs_e2ee_feature"]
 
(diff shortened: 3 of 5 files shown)


mulmarta (approved):

mls-rs/src/group/secret_tree.rs
```diff
@@ -269,6 +280,32 @@ impl<T: TreeIndex> SecretTree<T> {
 
         res
     }
+
+    /// Peeks at the next key generation, but does not increment the generation nor
+    /// derive keys.
+    ///
+    /// Takes &mut self since take_leaf_ratchet constructs and stores nodes in the
+    /// SecretTree the first time they are requested.
```
cindylindeed, line 288: Calling attention to this to verify it's true.

mls-rs/src/external_client/group.rs
```diff
@@ -638,6 +638,14 @@ where
         Ok(EventOrContent::Content(auth_content))
     }
 
+    #[cfg(all(feature = "export_key_generation", feature = "private_message"))]
+    async fn get_unauthenticated_key_generation_from_sender_data(
+        &mut self,
+        _cipher_text: &PrivateMessage,
+    ) -> Result<Option<u32>, MlsError> {
+        Ok(None)
```
cindylindeed, line 646: What are your thoughts on `Option<u32>` instead of `u32?` I thought `Option` encapsulated this external_client situation better, since it's not a MlsError, but I'm also not 100% convinced.

mls-rs/src/group/message_processor.rs
```diff
@@ -515,6 +524,22 @@ pub(crate) trait MessageProcessor: Send + Sync {
         #[cfg(feature = "by_ref_proposal")] cache_proposal: bool,
         time_sent: Option<MlsTime>,
     ) -> Result<Self::OutputType, MlsError> {
+        #[cfg(all(feature = "export_key_generation", feature = "private_message"))]
+        // For encrypted application messages, retrieve the unauthenticated key
+        // generation used to decrypt the message and return it with the plaintext. Does
+        // not return an error on failure, allowing `get_event_from_incoming_message` to
+        // continue owning that task.
```
cindylindeed, line 531: Calling attention to this (not returning an error) to verify you agree.

tomleavy:
Hey @cindylindeed can you take a look at the build failures. Also @mulmarta if you have a moment take a look at the callouts above

mulmarta (approved):
Thank you very much @cindylindeed for the solution!

PR #296, opened 2025-07-18
 Create new method to return inner dyn error

ethanfire:
In the previous implementation, attempts to downcast an AnyError by calling .source() would fail and return None because .source() was being called recursively. Example:
```
let error_with_data = CodecError::Custom(1);
let any_error: AnyError = error_with_data.into_any_error();

let downcasted_source = any_error
    .source()
    .downcast_ref::<CodecError>()
```
Previously, the downcasted_source would be None. Now, it will be successfully downcast to CodecError::Custom(1)


mls-rs-core/src/error.rs
```diff
@@ -23,7 +23,7 @@ impl Display for AnyError {
 #[cfg(feature = "std")]
 impl std::error::Error for AnyError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        self.0.source()
+        Some(&*self.0)
```
stefunctional, line 26: The drawback of this implementation is that if one prints an `AnyError` and its source recursively, the first message will be duplicated.

`AnyError` should be considered equivalent to its inner error. It does not add any layer of information, which is why its `Display` implementation just forwards to the inner error. As such, I think an inherent `downcast`  method or a method returning the inner `dyn Error` would make more sense. This is closer to what `anyhow::Error` does.

ethanfire: Ah thank you, that makes sense. I went ahead and added a new method to return the inner dyn error like you suggested.

tomleavy (dismissed):

mls-rs-core/src/error.rs
ethanfire: Updated the function, had to modify it to avoid an unknown-size-at-compile-time error

stefunctional: Looks good to me. Thank you!

stefunctional (approved):

tomleavy:
🍻

PR #299, opened 2025-08-11
Custom proposals support GSMA RCS encoding

emarteca:
The current implementation of the CustomProposal is that all user-specified custom proposals are wrapped in the CustomProposal struct. The user-specified custom proposal is encoded into a byte vector field in CustomProposal.

```
struct CustomProposal {
    proposal_type: ProposalType,
    data: Vec<u8>, // this is the mls_encoded user-specified custom proposal
}
```

This is very elegant as it means we don't need to extend the Proposal enum for every new proposal kind. However it also means that the encoding of CustomProposal is always as a byte vector. This means it is not compatible with other implementations that have custom proposals encoded directly (as top-level in the Proposal variants).

To ensure compatibility with other implementations of the [GSMA RCS spec](https://www.gsma.com/solutions-and-impact/technologies/networks/wp-content/uploads/2025/03/RCC.16-v1.0.pdf), this PR:
* adds the GSMA RCS spec-specific custom proposals as built-in ProposalType variants
* adds a new trait `CustomDecoder` that implements mapping to/from the `data` field in `CustomProposal`. This maps from the ProposalType to the encoding/decoding. For the GSMA RCS-specific proposal types, it will not encode the length of the `data` field, as it's not encoded as a `Vec<u8>`, but rather it will encode/decode directly as the GSMA RCS-specific proposal shapes:
    * EndMls: empty struct
    * RcsSignature: empty struct
    * ServerRemove: `struct { u32 }`, the same shape as the `RemoveProposal`
* hides all of this behind a new feature flag `gsma_rcs_e2ee_feature`, so the original functionality is preserved by default
* \+ tests

### DIFF
--- a/mls-rs-core/Cargo.toml
+++ b/mls-rs-core/Cargo.toml
@@ -23,6 +23,7 @@ serde = ["dep:serde", "zeroize/serde", "hex/serde", "dep:serde_bytes"]
 last_resort_key_package_ext = []
 post-quantum = []
 self_remove_proposal = []
+gsma_rcs_e2ee_feature = []
 
 [dependencies]
 mls-rs-codec = { version = "0.7", path = "../mls-rs-codec", default-features = false}
--- a/mls-rs-core/src/group/proposal_type.rs
+++ b/mls-rs-core/src/group/proposal_type.rs
@@ -58,6 +58,8 @@ impl ProposalType {
     pub const GROUP_CONTEXT_EXTENSIONS: ProposalType = ProposalType(7);
     #[cfg(feature = "self_remove_proposal")]
     pub const SELF_REMOVE: ProposalType = ProposalType(0xF003);
+    pub const RFC_SIGNATURE: ProposalType = ProposalType::new(0xF002);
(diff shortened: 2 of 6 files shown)


tomleavy:
Hey @emarteca thanks for looking into this, it's actually something that has been awkward for a while now because the serialization of CustomProposal makes it so that you can't implement future RFC proposals. 

One question I have here. Is it possible to integrate this into MlsCustomProposal so that you don't need to only turn this on for your feature? Maybe the MlsCustomProposal has a type declaration for the CustomDecoder and we have a DefaultCustomDecoder that is there by default which is the empty trait implementation you provided? I haven't actually played around with it to understand if this has the same generic pollution issue you described, but maybe possible.

emarteca:
Thanks for taking a look! I did originally try to go that route, but I couldn't figure out how to map to the correct decoder during the decoding. The issue I ran into is that we need a way to map from the custom proposal type to the decoder, since, when decoding the proposal, the only information we have is the proposal type.

mulmarta (approved):

PR #316, opened 2025-09-30
fix: improve edge case detection in HashMap and BTreeMap codec decoding

tomleavy:
### Description of changes:

This resolves two issues

1. HashMap and BTreeMap MlsDecode implementation allows for duplicate keys to be discarded, which should throw an error instead
2. HashMap and BTreeMap MlsDecode fails if presented with a specially crafted payload + zero length structure target for decode 

### Testing:

This file lacked proper tests, which have now been added

### DIFF
--- a/mls-rs-codec/src/lib.rs
+++ b/mls-rs-codec/src/lib.rs
@@ -50,6 +50,8 @@ pub enum Error {
     UnsupportedEnumDiscriminant,
     #[cfg_attr(feature = "std", error("Expected UTF-8 string"))]
     Utf8,
+    #[cfg_attr(feature = "std", error("Invalid map content"))]
+    InvalidMapContent,
     #[cfg_attr(feature = "std", error("mls codec error: {0}"))]
     Custom(u8),
 }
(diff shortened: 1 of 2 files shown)


stefunctional:
The implementation for `Vec<T>` suffers from the same problem when `T` is zero-sized, leading to an infinite loop for a malicious payload pretending to have a non-empty `Vec`.

stefunctional:
It looks like `git add mls-rs-codec/src/vec.rs`  might have been forgotten before the second commit.

mls-rs-codec/src/lib.rs
```diff
@@ -50,6 +50,8 @@ pub enum Error {
     UnsupportedEnumDiscriminant,
     #[cfg_attr(feature = "std", error("Expected UTF-8 string"))]
     Utf8,
+    #[cfg_attr(feature = "std", error("Invalid map content"))]
```
stefunctional, line 53: This error is not specific to maps anymore.

stefunctional (approved):

PR #320, opened 2025-10-01
fix: always use constant time comparisons for tag values

tomleavy:
### Description of changes:

Since `confirmation_tag` and `membership_tag` are MAC values, we should use constant time comparisons for them in all cases. This behavior is covered everywhere by custom implementing PartialEq with the subtle crate's `ct_eq` function

### Testing:

Covered by existing tests


mulmarta:
Should we do this for all structs with secrets that derive PartialEq, or remove PartialEq derivation? ? For example
* `KeySchedule`
* `HpkeSecretKey`
* `SignatureSecretKey`
* `EpochSecrets`
* `PreSharedKey`

stefunctional (approved):

PR #351, opened 2026-03-25
Add PSK-mode HPKE encrypt/decrypt API

### LINKED ISSUE #350: Add PSK-mode HPKE encrypt/decrypt API for targeted messages
### Description of feature:

Add PSK-mode HPKE encryption and decryption to the CipherSuiteProvider trait and the Group API. PSK-mode HPKE (RFC 9180, Section 5.1.2).

### Use case:

We want to be able to encrypt data to a given user with PSK for additional security.

If you are planning to implement this feature yourself, please give a brief overview of the changes
you would be making to accomplish it.

trebacz626:
### Issues:

Resolves #350 

### Description of changes:

Adds PSK-mode HPKE (RFC 9180) encryption and decryption support across the library.

### Call-outs:

* added new methods instead of adding optional Psk argument to not be a breaking change for sb who uses them, but can change
* hpke_seal_psk and hpke_open_psk are new required methods on CipherSuiteProvider, which is a breaking change for downstream trait implementors - can make them default not implemented 

### Testing:

Unit tests + integration tests in our codebase using this feature.


trebacz626:
I've added the infrastructure for hpke_psk_tests in test_suite.rs following the same pattern as the other tests (e.g., hpke_tests, signature_tests). However, I wasn't able to regenerate crypto_provider.json to include the new PSK test vectors, could you help me with that?

mulmarta (commented):
Thank you!

The test vector crypto_provider.json was generated using openssl in [this test](https://github.com/awslabs/mls-rs/blob/b2bb06c77b5116a35fc6b2dc35902f22bb3d5c78/mls-rs-crypto-openssl/src/lib.rs#L366). Which is definitely not the best way to do this...

mls-rs-core/src/psk.rs
```diff
@@ -110,6 +110,32 @@ impl From<Vec<u8>> for ExternalPskId {
...
+/// A pre-shared key value bundled with its identifier, for use with PSK-mode
+/// HPKE (RFC 9180).
+///
+/// Use [`PskBundle::new`] when no identifier is needed (empty `psk_id`),
+/// or [`PskBundle::new_with_id`] to set an explicit identifier.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PskBundle {
+    pub psk: PreSharedKey,
+    pub psk_id: ExternalPskId,
+}
```
mulmarta: Why not use the `Psk` struct from [hpke.rs](https://github.com/awslabs/mls-rs/blob/b2bb06c77b5116a35fc6b2dc35902f22bb3d5c78/mls-rs-crypto-hpke/src/hpke.rs#L77)? 

This module is about MLS's PSK which is something different - `ExternalPskId` is related to "external" MLS PSK versus e.g. "branching" MLS PSK. External PSK is provided by applications, and branching PSK is used internally by MLS.

trebacz626: I didn't want mls-rs-core to depend on mls-rs-crypto-hpke, but you are right, my current solution is not right either, I decided to move the Psk (renamed to HpkePsk) from mls-rs-crypto-hpke to mls-rs-core.

mulmarta (approved):

JHabz (approved):

PR #352, opened 2026-03-27
Bug in parent hash validation: skipping sibling tree hash recomputation for blank ancestors of nodes with changed unmerged_leaves

emarteca:
Bug: It is possible to construct a tree through sequential add/remove commits where the tree then fails to validate. This failure is in the parent hash validation, specifically in cases where during sibling tree hash computation, the sibling tree has a blank ancestor to a node that has a changed `unmerged_leaves` list.

It looks like the situation is as follows:
* on an incoming commit or join, the parent hashes must be validated
* as part of that validation, a set of original hashes is computed, to compare with. [here](https://github.com/awslabs/mls-rs/blob/main/mls-rs/src/tree_kem/parent_hash.rs#L186)
* during the computation of the original hashes, the hashes for a given subtree are only recomputed if the result of a helper function `different_unmerged` returns true [here](https://github.com/awslabs/mls-rs/blob/main/mls-rs/src/tree_kem/tree_hash.rs#L202)
* however, looking at `different_unmerged` [here](https://github.com/awslabs/mls-rs/blob/main/mls-rs/src/tree_kem/tree_hash.rs#L166), it appears there is a bug: if there is a subtree with a blank ancestor then the check will shortcircuit and return false. However, trees should be recomputed if there is a blank ancestor and a descendent  of that ancestor that has different `unmerged_leaves` .

@menonsamir isolated the bug and found the fix: thank you!! :pray: 

This PR contains @menonsamir 's bug fix, and a test that repro'd the previous validation failure (with `ParentHashMismatch`)

### DIFF
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -3943,6 +3943,54 @@ mod tests {
         let (_, _) = test_group.join("r").await;
     }
 
+    #[cfg_attr(not(mls_build_async), maybe_async::must_be_sync)]
+    #[maybe_async::test(not(mls_build_async), async(mls_build_async, crate::futures_test))]
+    async fn even_larger_group_fails_with_parent_hash_mismatch_blank_intermediate_nonblank_grandparent() {
+        let mut test_group = test_group_custom(
+            TEST_PROTOCOL_VERSION,
+            TEST_CIPHER_SUITE,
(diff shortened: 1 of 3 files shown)


mulmarta (approved):
Thank you so much!

trebacz626 (approved):

PR #356, opened 2026-04-29
mls-rs-crypto-webcrypto: support WorkerGlobalScope (use globalThis.crypto)

Vit-User:
## Summary

The WebCrypto provider currently hard-codes `web_sys::window()` in `get_crypto()` and `random_bytes()`, which returns `None` inside any Worker context. `SubtleCrypto` and `Crypto` are defined on `WorkerGlobalScope` and its subtypes (`DedicatedWorker`, `SharedWorker`, `ServiceWorker`) per the W3C Web Cryptography API spec, so the provider fails unnecessarily when callers want to keep MLS state inside a Worker for XSS-against-keys hardening.

## Change

This PR resolves crypto via `globalThis.crypto` (read through `js_sys::global()` + `Reflect::get` + `dyn_into::<web_sys::Crypto>`), which works identically in `Window`, `DedicatedWorkerGlobalScope`, `SharedWorkerGlobalScope`, and `ServiceWorkerGlobalScope` without branching on the global type.

- No new feature flags
- No `Cargo.toml` change (`Window` and `Crypto` web-sys features already cover what is needed)
- The error variant `CryptoError::WindowNotFound` is renamed to `CryptoError::WebCryptoUnavailable` since the failure now describes the actual condition. If maintainers prefer backwards compatibility, happy to keep `WindowNotFound` as a `#[deprecated]` alias - just say the word and I will push a follow-up commit.

## Test

Added a `wasm_bindgen_test_configure!(run_in_dedicated_worker)` integration test (`mls-rs-crypto-webcrypto/tests/worker.rs`) that exercises `signature_key_generate` and `random_bytes` inside a `DedicatedWorkerGlobalScope`, validating the new code path. Existing `Window`-context tests pass unchanged. The existing CI workflow (`wasm_build.yml`) picks the new test up automatically via `wasm-pack test --headless --chrome --release`.

## Motivation

Building an MLS-encrypted advice channel for a regulated NZ financial services PWA, with the protocol state kept inside a `DedicatedWorker` so a successful XSS in the page does not become a sign/decrypt oracle. Discovered the issue via two-tab end-to-end smoke testing of mls-rs 0.55 + mls-rs-crypto-webcrypto 0.15 in WASM and would prefer to land this upstream rather than ship a perma-fork.

## References

- [SubtleCrypto on MDN](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) - exposed on both `Window` and `WorkerGlobalScope`
- [W3C Web Cryptography API](https://www.w3.org/TR/WebCryptoAPI/) §2

## Test plan

- [x] `cargo build --target wasm32-unknown-unknown --release` (with `RUSTFLAGS=\"--cfg mls_build_async\"`) compiles clean
- [x] No remaining `web_sys::window()` references in the crate
- [ ] CI WASM workflow passes (`wasm_build.yml` runs the new worker integration test in headless Chrome)

### DIFF
--- a/mls-rs-crypto-webcrypto/src/lib.rs
+++ b/mls-rs-crypto-webcrypto/src/lib.rs
@@ -25,7 +25,7 @@ use mls_rs_crypto_hpke::{
 
 use mls_rs_crypto_traits::{AeadType, KdfType, KemId};
 
-use wasm_bindgen::JsValue;
+use wasm_bindgen::{JsCast, JsValue};
 use web_sys::SubtleCrypto;
 use zeroize::Zeroizing;
 
@@ -43,8 +43,8 @@ pub enum CryptoError {
     JsValue(String),
     #[error("Key has wrong length for cipher suite")]
     WrongKeyLength,
-    #[error("Window not found")]
-    WindowNotFound,
+    #[error("WebCrypto API unavailable in this context")]
+    WebCryptoUnavailable,
     #[error("Invalid signature")]
     InvalidSignature,
     #[error("Der encoding error {0}")]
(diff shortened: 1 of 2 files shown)


mulmarta (approved):
Thank you for the fix!

If you're planning to use the Web Crypto provider, keep in mind that it currently does not work with Firefox and Safari. The part that doesn't work is computing the DH public key from raw secret key bytes [1]---a non-standard crypto API required by MLS. I think that it _should_ work so we can file a bug to Mozilla. I'm not sure what we can do about Safari.

Separately, Mozilla has an origin trial for in-browser MLS support; I thought you may find it interesting. [2]

[1] https://github.com/awslabs/mls-rs/blob/main/mls-rs-crypto-webcrypto/src/ec/ecdh.rs#L113
[2] https://github.com/beurdouche/explainers/blob/main/MessagingLayerSecurity.md

PR #361, opened 2026-06-16
Validate that sender from prior epoch is still in the tree; do not return aliased sender if leaf index is reused

emarteca:
Prior to this PR, in a scenario where a client's leaf index is reused in the group (client is removed, a new client is added) then application message decryption and custom proposal validation incorrectly attribute the message as being sent from the new, aliased leaf.

With this change, we now:
- check the signing identity of the sender
- check the signing identity of the sender's leaf index in the current tree
- if these don't match, return a `MemberNotFound` because it means the member is no longer in the current tree

The test I added in this PR shows the new behaviour, where it now returns a `MemberNotFound` error, prior to this change it would attribute the messages in the test to Carol.

**Note**: a more invasive fix would be to change PriorEpoch to store the signing identities, but both for storage size reasons and for simplicity I decided to just return this error instead. LMK what you think.

### DIFF
--- a/mls-rs/src/group/mod.rs
+++ b/mls-rs/src/group/mod.rs
@@ -1453,7 +1453,6 @@ where
                 )
                 .await?;
 
-                #[cfg(feature = "prior_epoch_membership_key")]
                 validate_sender_signature_key_from_prior_epoch(
                     &self.state.public_tree,
                     &epoch.signature_public_keys,
@@ -1609,7 +1608,6 @@ where
                 )
                 .await?;
 
-                #[cfg(feature = "prior_epoch_membership_key")]
                 validate_sender_signature_key_from_prior_epoch(
                     &self.state.public_tree,
                     &epoch.signature_public_keys,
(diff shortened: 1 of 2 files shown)


mulmarta (approved):
Really great find! I'm going to approve this because it's a security issue.

Long-term, I can see a scenario where a group member rotates the public key for security, which would make the fix too strict.

mulmarta:
> Really great find! I'm going to approve this because it's a security issue.
> 
> Long-term, I can see a scenario where a group member rotates the public key for security, which would make the fix too strict.

Maybe part of our effort on the 1.x branch to externalize storage would be the real solution --- A commit does not mutate the group. Instead have something like
```rust
let epoch = my_storage.load_epoch(group_id, epoch_id)?;
// commit_effect has public past_epoch, new_epoch fields
let commit_effect = client.new_epoch_builder(epoch, commit_message)?.build()?;

// Receiving a message returns sender's index in the past epoch and applications can
// store whatever additional info they want about that epoch.
client.receive_message(&past_epoch, message)?;
my_storage.store_epoch(commit_effect.new_epoch, commit_effect.past_epoch)?;
```

JHabz (approved):

chinmayeebandal (approved):

PR #366, opened 2026-07-20
Allow creating groups with an arbitrary start epoch (#327)

gildas-amazon:
Add a `custom_start_epoch` cargo feature that lets a locally created
group start at a configured epoch instead of 0. The epoch is set via a
new `ClientBuilder::start_epoch` builder option and applied to the
initial GroupContext in Client::create_group / create_group_with_id.

This is not RFC 9420 compliant (which mandates epoch 0) and is gated
behind the feature. It only affects groups created locally; groups
joined via Welcome or external commit adopt their epoch as usual, and
branch/re-init subgroups still start at epoch 0 (their Welcome epoch
must be 1).

The value is threaded through Settings and the ClientConfig trait
(defaulting to 0), including the blanket MlsConfig -> ClientConfig
forwarding so it is not silently dropped for real clients.

### Issues:

Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes:

Describe the current behavior and how your code changes that behavior. If there are no issues that this PR resolves, please create one first.

### Call-outs:

Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:

How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?


mulmarta (dismissed):
Makes sense.

It works as is, but in principle should epoch be a Config property, or an input to `create_group`? It may be easier with my PR #367 that replaces the many inputs to `create_group` with a builder.

gildas-amazon:
Yes make sense. I think it would be better to implement this on top of PR https://github.com/awslabs/mls-rs/pull/367

mulmarta:
@gildas-amazon #367 was merged.

mulmarta (approved):

mulmarta (approved):

PR #369, opened 2026-07-29
feat: Expose the UpdatePath leaf node of a commit on MlsMessage

### LINKED ISSUE #368: Expose the UpdatePath leaf node of a commit for delivery-service inspection
### Description of feature:

An accessor on `MlsMessage` for the leaf node carried in a commit's `UpdatePath`, e.g.

```rust
/// The leaf node in this commit's `UpdatePath`, if it has one.
pub fn commit_path_leaf_node(&self) -> Option<&LeafNode>;
```

returning `None` for non-commits and for commits without a path. A narrower
`Option<&SigningIdentity>`, or a field on `MlsMessageDescription`, would cover our case equally well.

### Use case:

We run a delivery service that inspects the public framing of `PublicMessage` commits with
[`description()` and `proposals_by_value()`](https://docs.rs/mls-rs/0.55.2/mls_rs/struct.MlsMessage.html).
A leaf that joins via an Add proposal is reachable there (`AddProposal::key_package()`, #165), so we
can check the credential of every joiner on that path. A leaf that joins via external commit is
not: it sits in the commit's `UpdatePath`, and
[`MlsMessagePayload` is `pub(crate)`](https://github.com/awslabs/mls-rs/blob/0.55.2/mls-rs/src/group/framing.rs#L618),
so no public API reaches it. So the one join path a DS can't vet is the external commit.

The data is public: the `UpdatePath` leaf in a `PublicMessage` is unencrypted, so reading it needs
no keys and no group state.

Alternatives we looked at: `ExternalGroup` does validate the path leaf, but it processes the commit
and tracks group state, which our stateless DS deliberately doesn't do. Hand-decoding the
`UpdatePath` with `mls-rs-codec` works, but then client and server no longer parse through the same
code.

I have a small PR ready: an accessor next to `proposals_by_value()` in
[`framing.rs`](https://github.com/awslabs/mls-rs/blob/0.55.2/mls-rs/src/group/framing.rs) plus a
test. Also fine with a different shape if you'd rather not expose the whole `LeafNode`.

7331:
### Issues:

Resolves #368

### Description of changes:

The leaf node carried in a commit's `UpdatePath` was not reachable through any public API, so a
delivery service inspecting the public framing of a `PublicMessage` commit could read the
`KeyPackage` of a member joining via an Add proposal but not the leaf of a member joining via an
external commit. This adds `MlsMessage::commit_path_leaf_node()`, mirroring the shape of
`proposals_by_value()`: `None` for anything that is not a plaintext commit carrying a path.

### Call-outs:

`LeafNode` is already part of the public API (re-exported as `mls_rs::group::LeafNode`), so no new
types are exposed. Happy to narrow the return to `&SigningIdentity` if you prefer.

### Testing:

New unit test builds a real external commit and asserts the accessor returns its leaf node (and
`None` for a non-commit message). Ran the full crate test suite, clippy (`--all-features` and the
bare-bones feature lane), and `cargo fmt --check`.


mulmarta:
Thanks @7331 it makes sense. I think returning the LeafNode is fine. Looks like clippy with no-std feature is failing (please ignore the slack notifier failure, we're fixing this).

7331:
Thanks! Added the same `#[allow(unreachable_patterns)]` that `proposals_by_value` already carries. With `private_message`/`by_ref_proposal` off, `Content` only has the `Commit` variant left, so the wildcard arm is dead in the no-std lane. `cargo +stable clippy --no-default-features -p mls-rs -- -D warnings` passes locally now.

mulmarta (approved):

chinmayeebandal (approved):

PR #380, opened 2026-08-13
feat: Expose a commit's by-reference proposals on MlsMessage

### LINKED ISSUE #377: Expose a commit's by-reference proposals
### Description of feature:

An accessor on `MlsMessage` for the references of a commit's by-reference proposals, e.g.

```rust
/// The `ProposalRef` of every proposal this commit carries by reference.
pub fn proposals_by_reference(&self) -> Vec<&ProposalRef>;
```

returning an empty list for non-commits and for all-inline commits. A plain count would cover our
case too.

### Use case:

A commit can carry a proposal itself ("by value") or just the hash of one published earlier ("by
reference"). `proposals_by_value()` hands over the first kind; `Commit` and `ProposalOrRef` are
`pub(crate)`, so there is no public way to see the second at all.

We run a delivery service that authorizes a commit by its proposal set, read through
[`proposals_by_value()`](https://docs.rs/mls-rs/0.55.2/mls_rs/struct.MlsMessage.html). A commit
whose operations are all by reference reports an empty proposal set, so "this route permits no
proposals" would wave through a remove or an add the server never saw.

The data is public: the proposal list of a `PublicMessage` commit is unencrypted.

Alternatives we looked at: `ExternalGroup::get_cached_proposals()` (#340) resolves references, but
it processes every proposal and tracks group state, which our stateless DS deliberately doesn't do.
So today we hand-decode the proposal vector with `mls-rs-codec` and cross-check the result against
what mls-rs reports for the same bytes, purely to catch layout drift. It works, but client and
server no longer parse through the same code (same situation as #368), and we would happily delete
all of it the day this accessor exists.

I have a small PR ready: an accessor next to `proposals_by_value()` in
[`framing.rs`](https://github.com/awslabs/mls-rs/blob/main/mls-rs/src/group/framing.rs) plus a
test, gated on `by_ref_proposal` since `ProposalRef` only exists there.

7331:
### Issues:

Resolves #377

### Description of changes:

A commit names its proposals by value or by reference. `MlsMessage::proposals_by_value()` returns
the first kind; `Commit` and `ProposalOrRef` are `pub(crate)`, so no public API reaches the second,
not even a count. A commit whose operations are all by reference then reports an empty proposal
set, which a delivery service authorizing by proposal set cannot distinguish from a plain key
rotation.

This adds `MlsMessage::proposals_by_reference()`, mirroring the shape of `proposals_by_value()`:
the `ProposalRef` of every by-reference entry for a plaintext commit, empty for anything else.

### Call-outs:

`ProposalRef` is already public under `by_ref_proposal` (`mls_rs::mls_rules::ProposalRef`), so no
new types are exposed; the accessor is gated on the same feature. `ExternalGroup::get_cached_proposals()`
(#340) covers resolving references for a service that tracks group state; a stateless DS inspecting
only public framing can't use it, so today we hand-decode the proposal vector ourselves and this
lets us delete that. Happy to narrow the return to a count if you prefer.

### Testing:

New unit test commits a cached proposal and asserts the accessor returns its ref, matching
`into_proposal_reference()` of the proposal message (and empty for an all-by-value commit and a
non-commit message). Ran the full crate test suite, clippy (`--all-features` and the bare-bones
feature lane), and `cargo fmt --check`.

---
_Recreation of #378, which was auto-closed when my fork was accidentally made private (permanently detaching it from the fork network). Identical commits (head df34e12)._

### DIFF
--- a/mls-rs/src/client.rs
+++ b/mls-rs/src/client.rs
@@ -931,6 +931,11 @@ mod tests {
     use assert_matches::assert_matches;
 
     #[cfg(feature = "by_ref_proposal")]
+    use crate::crypto::test_utils::test_cipher_suite_provider;
+    #[cfg(feature = "by_ref_proposal")]
+    use crate::key_package::test_utils::test_key_package_message;
+
+    #[cfg(feature = "by_ref_proposal")]
     use crate::group::message_processor::ProposalMessageDescription;
     #[cfg(feature = "by_ref_proposal")]
     use crate::group::proposal::Proposal;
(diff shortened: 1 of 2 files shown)


mulmarta (approved):

7331:
hi @mulmarta pinging on this too :)

chinmayeebandal (approved):

7331:
thank you!

we seem to be hitting the 30s chrome driver timeout, not sure what we can do other than retrying when convenient/it passes

PR #372, opened 2026-08-06
External group exported tree

andybug:
### Issues:

No associated issue.

### Description of changes:

`ExternalGroup::export_tree()` always MLS-encodes the whole ratchet tree to bytes, forcing callers who only want to inspect
tree structure (e.g. counting blank nodes) through an unnecessary encode/decode round trip. `Group` already exposes a
zero-copy `export_tree()` returning a borrowed `ExportedTree`; this PR adds the same capability to `ExternalGroup` under
the name `exported_tree()` (to avoid colliding with the existing bytes-returning method).

### Call-outs:

None — this is an additive, non-breaking API change (new method, no existing behavior modified).

### Testing:

Added a unit test (`external_group_exported_tree_matches_export_tree_bytes`) that:
- Builds a tree with a mix of occupied and blank leaves (adds carol, then removes bob).
- Asserts the new zero-copy `exported_tree()` matches the byte-encoded `export_tree()` output via `to_bytes()`.
- Asserts `get_leaf` correctly reflects blank vs. occupied leaves.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 and MIT license.

PR #381, opened 2026-08-13
feat: Expose the leaf slot count of an ExportedTree

7331:
### Issues:

No associated issue.

### Description of changes:

Reading the leaf slot count of an `ExportedTree` currently costs either a full copy of the
tree (`NodeVec::from(tree).total_leaf_count()` consumes it and calls `into_owned()`) or
re-deriving `ceil(len / 2)` from `nodes()`, which asks callers to know that MLS stores
leaves at even indices. This forwards the existing `NodeVec::total_leaf_count()` through the
borrow the tree already holds, giving the bound for `get_leaf()` with no allocation.

Same motivation as #372: the data is right there, only the borrow was missing.

`occupied_leaf_count()` is deliberately not forwarded — it is gated on
`custom_proposal + tree_index`, and `roster().members_iter().count()` already answers it.

### Testing:

Extended `test_exported_tree_accessors` over the existing 4-slot test tree: asserts the count,
that the last valid `LeafIndex` resolves, that one past it errors, and that the roster reports
3 occupied. Ran the full crate test suite, clippy (`--all-features` and the bare-bones lane),
and `cargo fmt --check`.

---
_Recreation of #379, which was auto-closed when my fork was accidentally made private (permanently detaching it from the fork network). Identical commits (head d6356b4)._

### DIFF
--- a/mls-rs/src/group/exported_tree.rs
+++ b/mls-rs/src/group/exported_tree.rs
@@ -96,6 +96,15 @@ impl<'a> ExportedTree<'a> {
         Ok(parent)
     }
 
+    /// Returns the number of leaf slots in the tree, counting blanks.
+    ///
+    /// Leaf indices run from `0` to `total_leaf_count() - 1`, so this is the
+    /// bound for [`get_leaf`](Self::get_leaf). Use [`roster`](Self::roster)
+    /// to iterate the occupied leaves only.
+    pub fn total_leaf_count(&self) -> u32 {
+        self.0.total_leaf_count()
+    }
+
     /// Returns the leaf node at the given `LeafIndex`, or `None` if the
     /// leaf slot is blank. Returns an error if the index is out of range.
     pub fn get_leaf(&self, index: LeafIndex) -> Result<Option<&LeafNode>, MlsError> {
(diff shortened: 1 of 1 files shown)


mulmarta (approved):
I don't know the exact use case so just to make sure you're aware - this outputs the next power of 2 after `ceil(nodes().len() / 2)`. For example for a tree with leaves 0, 1, 2, 3, 4, all full, `total_leaf_count` outputs 8 even though `nodes().len()` is 9. The reason is that a ratchet tree is always full but the trailing blanks aren't stored. (Also the update path in a commit from leaf 4 always contains only the leaf and the root because "useless" nodes  in the direct path are skipped.)

7331:
> I don't know the exact use case so just to make sure you're aware - this outputs the next power of 2 after `ceil(nodes().len() / 2)`. For example for a tree with leaves 0, 1, 2, 3, 4, all full, `total_leaf_count` outputs 8 even though `nodes().len()` is 9. The reason is that a ratchet tree is always full but the trailing blanks aren't stored. (Also the update path in a commit from leaf 4 always contains only the leaf and the root because "useless" nodes in the direct path are skipped.)

Thank you for the response!

My delivery service parses the tree out of a published GroupInfo and needs the bound on addressable leaf indices, which slots *can* exist, before separately checking occupancy against the roster/credentials.

The membership count could/would be the wrong number there. I believe "blanks" are included at this point.

Should I update the the PR with the 5 -> 8 example?

7331:
hey @mulmarta i hope all is fine on your end.

I came back to check in on this PR, but I see that nobody else has been active here except you.

Let me know if i can help, this is is a passion project of mine. it's fun for me!

Im looking forward to it!
Thanks!

(ps: wasm failed, but it was a timeout, pls re-run the workflows)

mulmarta (approved):

chinmayeebandal (approved):

